PR #2, opened 2023-10-18
Add print_dictionary helper

Hannibal404:


Hannibal404:
Made the organization membership public

drgn_tools/table.py
```diff
@@ -42,3 +43,37 @@ def print_table(
...
+    out = sys.stdout
+    if outfile and report:
+        out = open(outfile, "a")
+        print("\n", file=out)
+    elif outfile:
+        out = open(outfile, "w")
+
+    for title in dictionary:
+        print(
+            f"{title.ljust(lcol_length)}: {dictionary[title]:<}",
```
brenns10, line 74: `{dictionary[title]:<}`
nit: technically, left alignment is the default, so there's no need to specify it

brenns10 (changes requested):
One small nit but otherwise, it looks good to me!

brenns10 (approved):

brenns10:
Thanks for the quick work unblocking both pull requests!

PR #4, opened 2023-10-20
Add cpuinfo module

Hannibal404:


drgn_tools/cpuinfo.py
```diff
@@ -0,0 +1,615 @@
...
+            flags >>= 1
+
+    for i in range(nbugints):
+        flags = int(cpu_caps_bugs[i + ncapints])
+        for j in range(32):
+            if (flags & 1) and bug_flags[i * 32 + j]:
+                bug_flags_list.append(
+                    bug_flags[i * 32 + j].string_().decode("utf-8")
+                )
+            flags >>= 1
```
brenns10, line 74: I think we can simplify this substantially. A few notes:

1. Drgn has a `for_each_set_bit()` helper that will make the loops much nicer.
2. On the master branch of Linux (and indeed, OL9 RHCK), `NBUGINTS` is already 2. We can't hard-code that value. Instead, let's infer it from the size of `x86_bug_flags`. In fact... we don't really even need to infer those sizes directly!
3. Let's also use `read_()` here to avoid repeated re-reading of the array data.

```suggestion
    cap = cpuinfo_struct.x86_capability
    cap_flags = prog["x86_cap_flags"].read_()
    bug_flags = prog["x86_bug_flags"].read_()

    cpu_flags_list = []
    bug_flags_list = []

    for nr in for_each_set_bit(cap, 8 * sizeof(cap)):
        if nr < len(cap_flags) and cap_flags[nr]:
            cpu_flags_list.append(cap_flags[nr].string_().decode("utf-8"))
        elif nr >= len(cap_flags) and bug_flags[nr - len(cap_flags)]:
            bug_flags_list.append(bug_flags[nr - len(cap_flags)].string_().decode("utf-8"))
```

drgn_tools/cpuinfo.py
```diff
@@ -0,0 +1,615 @@
...
+    # `struct cpuinfo_arm64` being fetched using 'boot_cpu_data' only has
+    # register values, while the struct fetched using 'cpu_data' has all values
+    #  zero/null
+
+    cpus = int(prog["nr_cpu_ids"])
+
+    if "cpu_data" in prog:
+        cpuinfo_struct = prog["cpu_data"]
+    elif "boot_cpu_data" in prog:
+        cpuinfo_struct = prog["boot_cpu_data"]
```
brenns10, line 34: Let's use `.read_()` here ([docs](https://drgn.readthedocs.io/en/latest/api_reference.html#drgn.Object.read_)).

The reasoning here is that the cpuinfo struct is a large-ish structure, and we're going to be reading parts of it very frequently. The way drgn handles this (in the live case) is by calling `read(/proc/kcore)` for each time we access a value. And reading from /proc/kcore has a quite high overhead for each call (but the overhead for each additional byte of data read is much lower). Thus, it makes sense to read chunks of data at a time, rather than small reads. If we call `.read_()` like shown, then the whole structure gets read at once, and the result is saved as the value of the object.

This means that the value won't be updated to reflect changes over time. But that also is good, because the chances are much better that your read would be atomic, in the live case.
```suggestion
    if "cpu_data" in prog:
        cpuinfo_struct = prog["cpu_data"].read_()
    elif "boot_cpu_data" in prog:
        cpuinfo_struct = prog["boot_cpu_data"].read_()
```

brenns10 (commented):
I'm sorry I haven't gotten the chance to do a detailed review of the PR yet, but I did get started on it. Here's some comments for the first few functions. I'll send more feedback tomorrow!

drgn_tools/cpuinfo.py
```diff
@@ -0,0 +1,615 @@
...
+    """
+    Extracts Mitigation for Spectre_V2
+    """
+    if "SPECTRE_V2_LFENCE" not in prog:
+        mitigation = (
+            prog["spectre_v2_strings"][prog["spectre_v2_enabled"]]
+            .string_()
+            .decode("utf-8")
+        )
+        if test_taint(prog, 16) and (
```
brenns10, line 170: We have the taint flags defined in `drgn_tools.taint`:

```
from drgn_tools.taint import Taint
from drgn.helpers.linux.bitops import test_bit

test_bit(Taint.AUX, prog["taint_flags"].address_of_())
```

drgn_tools/cpuinfo.py
```diff
@@ -0,0 +1,615 @@
...
+            mitigation = (
+                "Vulnerable: eIBRS+LFENCE with unprivileged eBPF and SMT"
+            )
+        else:
+            mitigation = (
+                prog["spectre_v2_strings"][prog["spectre_v2_enabled"]]
+                .string_()
+                .decode("utf-8")
+            )
+            if (cpu_caps_bugs[7] >> 26) & 1:
```
brenns10, line 201: Please define a constant and use `test_bit()`

drgn_tools/cpuinfo.py
```diff
@@ -0,0 +1,615 @@
...
+    Extracts Mitigation for MDS
+    """
+    mitigation = (
+        prog["mds_strings"][prog["mds_mitigation"]].string_().decode("utf-8")
+    )
+    mitigation += "; SMT "
+    nbugints = 1
+    ncapints = len(cpu_caps_bugs) - nbugints
+    if (cpu_caps_bugs[4] >> 31) & 1:
+        mitigation += "Host state unknown"
```
brenns10, line 303: constant + test_bit

drgn_tools/cpuinfo.py
```diff
@@ -0,0 +1,615 @@
...
+    elif (cpu_caps_bugs[ncapints] >> 20) & 1:
```
brenns10, line 304: constant + test_bit

drgn_tools/cpuinfo.py
```diff
@@ -0,0 +1,615 @@
...
+    mitigation = (
+        prog["taa_strings"][prog["taa_mitigation"]].string_().decode("utf-8")
+    )
+
+    if (
+        prog["taa_mitigation"] == prog["TAA_MITIGATION_TSX_DISABLED"]
+        or prog["taa_mitigation"] == prog["TAA_MITIGATION_OFF"]
+    ):
+        pass
+    elif (cpu_caps_bugs[4] >> 31) & 1:
```
brenns10, line 333: constant + test_bit

drgn_tools/cpuinfo.py
```diff
@@ -0,0 +1,615 @@
...
+
+def get_itlb_multihit_mitigation(prog: Program, cpu_caps_bugs: Object) -> str:
+    """
+    Extracts Mitigation for itlb_multihit
+    """
+    if "l1tf_vmx_states" not in prog:
+        mitigation = "Processor vulnerable"
+    elif "cr4_read_shadow" in prog:
+        if (not ((cpu_caps_bugs[7] >> 31) & 1)) or (
+            not ((cpu_caps_bugs[4] >> 5) & 1)
```
brenns10, line 353: constant + test_bit

drgn_tools/cpuinfo.py
```diff
@@ -0,0 +1,615 @@
...
+
+def get_mmio_stale_data_mitigation(
+    prog: Program, cpu_caps_bugs: Object
+) -> str:
+    """
+    Extracts Mitigation for mmio_stale_data and mmio_unknown
+    """
+    nbugints = 1
+    ncapints = len(cpu_caps_bugs) - nbugints
+    if (cpu_caps_bugs[ncapints] >> 26) & 1:
```
brenns10, line 385: constant + test_bit

Since this is a "bug" field, which means you need to determine "ncapints" first, be sure to detect ncapints correctly via `len(prog["x86_cap_flags"]) / 32`, rather than hard-coding nbugints to 1 as you have here.

drgn_tools/cpuinfo.py
```diff
@@ -0,0 +1,615 @@
...
+        mitigation = "Unknown: No mitigations"
+    else:
+        mitigation = (
+            prog["mmio_strings"][prog["mmio_mitigation"]]
+            .string_()
+            .decode("utf-8")
+        )
+        if prog["mmio_mitigation"] == prog["MMIO_MITIGATION_OFF"]:
+            pass
+        elif (cpu_caps_bugs[4] >> 31) & 1:
```
brenns10, line 395: constant + test_bit

drgn_tools/cpuinfo.py
```diff
@@ -0,0 +1,615 @@
...
+        "Retbleed",
+        "spec_store_bypass",
+        "GDS",
+        "SRSO",
+    ]
+
+    if "cpu_data" in prog:
+        cpuinfo_struct = prog["cpu_data"]
+    elif "boot_cpu_data" in prog:
+        cpuinfo_struct = prog["boot_cpu_data"]
```
brenns10, line 512: Let's use `read_()` here again

drgn_tools/cpuinfo.py
```diff
@@ -0,0 +1,615 @@
...
+        "tsx_async_abort",
+        "itlb_multihit",
+        "SRBDS",
+        "mmio_stale_data",
+        "mmio_unknown",
+        "Retbleed",
+        "spec_store_bypass",
+        "GDS",
+        "SRSO",
+    ]
```
brenns10, line 507: Let's turn this into a dictionary mapping the vulnerability to the checking function, and add another table mapping the vulnerability to the CPU bug name, if it's different. This way, you don't need that huge if ... elif ... elif block in the loop, you can just lookup the relevant function from the dict.

You'll need to make sure that all of these `get_xxxx_checker()` functions accept two args: `(prog, cpu_caps_bugs)`. A few of them just take `prog`, so you'll need to add a second argument for this to work properly.

Please also create a `get_srso_mitigation()` function that simply returns `"Vulnerable (Status Unknown)"`, so that you don't mix that logic up with this function, which should be simply about running the vulnerability checker for each CPU bug.

The benefit of all of this is that when the next CPU vulnerability happens, we just add a checker, and update each table, and we don't need to make sure we're updating that big if-statement.

```suggestion
    vuln_to_checker = {
        "Meltdown": get_meltdown_mitigation,
        "Spectre_V1": get_spectre_v1_mitigation,
        "Spectre_V2": get_spectre_v2_mitigation,
        "L1TF": get_l1tf_mitigation,
        "MDS": get_mds_mitigation,
        "tsx_async_abort": get_taa_mitigation,
        "itlb_multihit": get_itlb_multihit_mitigation,
        "SRDBS": get_srbds_mitigation,
        "mmio_stale_data": get_mmio_stale_data_mitigation,
        "mmio_unknown": get_mmio_stale_data_mitigation,
        "Retbleed": get_retbleed_mitigation,
        "spec_store_bypass": get_ssb_mitigation,
        "GDS": get_gds_mitigation,
        "SRSO": GET_SRSO_MITIGATION,
        "SSBD": None,
    }
    bug_to_vuln_name = {
        "cpu_meltdown": "Meltdown",
        "spectre_v1": "Spectre_V1",
        "spectre_v2": "Spectre_V2",
        "l1tf": "L1TF",
        "mds": "MDS",
        "taa": "tsx_async_abort",
        "srbds": "SRBDS",
        "retbleed": "Retbleed",
        "gds": "GDS",
        "srso": "SRSO",
    }
```

drgn_tools/cpuinfo.py
```diff
@@ -0,0 +1,615 @@
...
+        elif bug == "gds":
+            mitigations["GDS"] = get_gds_mitigation(prog)
+
+        elif bug == "srso":
+            mitigations["SRSO"] = "Vulnerable (Status Unknown)"
+            # Currently cannot handle this mitigation data because the kernel
+            # uses assembly code to fetch it
+
+        else:
+            mitigations[bug] = "Vulnerable"
```
brenns10, line 585: Assuming you made the hash table of mitigations, then you can just do the following:

```suggestion
    mitigations = {vuln: "Not Affected" for vuln in vuln_to_checker}
    for bug in bugs:
        vuln_name = bug_to_vuln_name.get(bug, bug)
        mitigation_checker = vuln_to_checker.get(vuln_name)
        if mitigation_checker:
            mitigations[vuln_name] = mitigation_checker(prog, cpu_caps_bugs)
        else:
            mitigations[vuln_name] = "Vulnerable"
```

brenns10 (changes requested):
Ok, finished reviewing this. Thanks for doing all of the logic for these different CPU mitigations. I know there's a lot of it here.

drgn_tools/cpuinfo.py
Hannibal404: What is the significance of `"SSBD": None` in vuln_to_checker? Because I don't remember encountering any such bug.

brenns10: Your `vulns` list contains `SSBD`, just after `Spectre_V2`. Later on you create the `mitigations` based on the list, so the `mitigations` dictionary you return ends up containing `SSBD: Not Impacted` (since there's no check in the if/elif/elif block that updates it).

If that wasn't intentional, then don't include it there. I was just trying to keep my suggestion identical to yours in logic.

Hannibal404: Ah I see. It was there in the sample I was seeing in the JIRA, but there were no functions for it in the kernel. I'll remove it. Thanks!

drgn_tools/cpuinfo.py
```diff
@@ -0,0 +1,581 @@
...
+
+
+def get_mmio_stale_data_mitigation(
+    prog: Program, cpu_caps_bugs: Object
+) -> str:
+    """
+    Extracts Mitigation for mmio_stale_data and mmio_unknown
+    """
+    ncapints = len(prog["x86_cap_flags"]) // 32
+    if test_bit(ncapints * 32 + X86_BUG_MMIO_UNKNOWN):
```
brenns10, line 374: I guess when you write it like that, there's no reason to divide by 30 :)
Also, I think you forgot to add `cpu_caps_bugs` to the argument.
```suggestion
    num_caps = len(prog["x86_cap_flags"])
    if test_bit(num_caps + X86_BUG_MMIO_UNKNOWN, cpu_caps_bugs):
```

Hannibal404: The division is actually needed, because each bit in `cpu_caps_bugs` corresponds to an element of `x86_cap_flags`, so it has `num_caps * 32` elements.

drgn_tools/cpuinfo.py
```diff
@@ -0,0 +1,581 @@
...
+
+X86_CR4_VMXE = 1 << 13
+
+X86_VENDOR_HYGON = 2
+X86_VENDOR_AMD = 3
+
+SPEC_CTRL_IBRS_FIRMWARE = 1 << 3
+SPEC_CTRL_IBPB_INUSE = 1 << 0
+
+TAINT_NO_RETPOLINE = 16
```
brenns10, line 41: I searched high and low for where you possibly could have come from with this. It appears in [this](https://lore.kernel.org/all/20180112175507.31750-1-andi@firstfloor.org/) patch series but in no upstream kernel, ever.

It turns out that the series there got included into UEK4 for some reason...

At the point of the code where you use this, please do write a nice comment explaining what this is, which kernel versions it's valid for, and why it exists. As it is, taint #16 is `TAINT_AUX` in any normal kernel, so this is highly unusual.

brenns10 (changes requested):
One last thing here, then I think we'll be good, assuming tests pass.

brenns10:
Looking good, but there's an internal CI failure:
```

    def get_retbleed_mitigation(prog: Program, cpuinfo: Object) -> str:
        """
        Extracts Mitigation for Retbleed
        """
        if "retbleed_state" in prog:
            if prog["retbleed_state"] == prog.constant(
                "RETBLEED_MITIGATION_UNRET"
            ):
                if (
>                   cpuinfo.x86_vendor != X86_VENDOR_AMD
                    and cpuinfo.x86_vendor != X86_VENDOR_HYGON
                ):
E               AttributeError: '_drgn.Object' object has no attribute 'x86_vendor'
```

Looks like `get_retbleed_mitigation()` expects the cpuinfo object, not the feature/bug flags bitfield. You could add that as another parameter or just look it up from `prog`.

brenns10 (approved):
This is looking good, thanks!

PR #5, opened 2023-10-24
meminfo: Add meminfo corelens module

jianfenw:
This commit implements a meminfo corelens module that enables users to dump detailed statistics of the memory management subsystem. The output is similar to 'cat /proc/meminfo', which includes many aspects of the mm subsystem.

This module supports UEK 5, 6, and 7 and for both x86-64 and aarch64. It is tested for all these above settings. For each case, this meminfo module's output is compared against the output of `cat /proc/meminfo`. Results match closely with only small differences.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -119,3 +119,10 @@ drgn_tools.cmdline
 
 .. automodule:: drgn_tools.cmdline
    :members:
+
+
+drgn_tools.meminfo
+-----------------------
+
+.. automodule:: drgn_tools.meminfo
+   :members:
--- a/drgn_tools/meminfo.py
+++ b/drgn_tools/meminfo.py
@@ -1,0 +1,654 @@
+# Copyright (c) 2023, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
+"""
+Helpers for dumping memory usage information and statistics
+"""
+import argparse
+import math
+from collections import UserDict
+from typing import Dict
+from typing import Iterator
+from typing import List
+
+import drgn
+from drgn import Object
+from drgn import Program
+from drgn.helpers.linux import list_for_each_entry
+from drgn.helpers.linux.percpu import percpu_counter_sum
+
+from drgn_tools.corelens import CorelensModule
+from drgn_tools.mm import totalram_pages
+from drgn_tools.util import has_member
+
+
+__all__ = ("show_all_meminfo", "get_all_meminfo")
+
+
+class StatDict(UserDict):
+    """A dictionary wrapper class to handle renaming of statistics items."""
+
+    renames = {
+        "NR_SLAB_RECLAIMABLE": "NR_SLAB_RECLAIMABLE_B",
+        "NR_SLAB_UNRECLAIMABLE": "NR_SLAB_UNRECLAIMABLE_B",
+        "NR_ANON_MAPPED": "NR_ANON_PAGES",
+        "NR_KERNEL_STACK_KB": "NR_KERNEL_STACK",
+    }
+
+    def __getitem__(self, key):
+        """
+        Some global statistics items's names was changed in the past. This
+        overwrites the default __getitem__ function to correct name for these.
+        """
+        if key in self.renames and key not in self.data:
+            key = self.renames[key]
+        return super().__getitem__(key)
+
+
+def _read_stats(prog: Program, arr_name: str, enum: str) -> UserDict:
+    """
+    Read counters from a global statistics array with name ``arr_name``.
+    Parse all statistics items that are indexed by ``enum`` and store results
+    in a dictionary.
+
+    :param prog: drgn program
+    :param arr_name: The global statistics array's variable name
+    :param enum: The global statistics array's corresponding enum type
+    :returns: A dictionary that contains all statistics items in the array
+    """
+    arr = prog[arr_name].read_()
+    enum_obj = prog.type(enum)
+    stats = StatDict()
+    # Skip the last item, which is the length of the array
+    for name, value in enum_obj.enumerators[:-1]:
+        stats[name] = max(0, arr[value].counter.value_())
+    return stats
+
+
+def get_global_mm_stats(prog) -> UserDict:
+    """
+    Read counters from ``vm_zone_stat``, ``vm_node_stat``, and ``vm_numa_stat``
+    for global zone/node page statistics. All statistics values are stored in
+    a customized dictionary, as this function's return.
+
+    :param prog: drgn program
+    :returns: A dictionary that contains all global statistics items
+    """
+    try:
+        stats = _read_stats(prog, "vm_zone_stat", "enum zone_stat_item")
+    except KeyError:
+        stats = _read_stats(prog, "vm_stat", "enum zone_stat_item")
+    try:
+        node_stat = _read_stats(prog, "vm_node_stat", "enum node_stat_item")
+        stats.update(node_stat)
+    except KeyError:
+        pass
+    try:
+        numa_stat = _read_stats(prog, "vm_numa_stat", "enum numa_stat_item")
+        stats.update(numa_stat)
+    except KeyError:
+        pass
+    return stats
+
+
+def get_mm_constants(prog: Program) -> Dict[str, int]:
+    """
+    Parse the systemwise and architecture-specific parameters that are relevant
+    to the kernel's memory subsystem. Results are cached in a dictionary.
+
+    :param prog: drgn program
+    :returns: A dictionary that contains relevant systemwise parameters
+    """
+    cache = prog.cache.setdefault("drgn_tools.meminfo", {})
+    mm_consts = cache.get("mm_constants", {})
+    if not mm_consts:
+        mm_consts = {}
+
+        # Determine max numbers of NUMA nodes and memory zones in a node.
+        # In mm/mempolicy.c, ``preferred_node_policy`` is defined as the following:
+        # "static struct mempolicy preferred_node_policy[MAX_NUMNODES];"
+        # Also, ``MAX_NUMNODES`` equals to (1 << NODES_SHIFT).
+        mm_consts["MAX_NUM_NODES"] = len(prog["preferred_node_policy"])
+        mm_consts["NODES_SHIFT"] = int(math.log2(mm_consts["MAX_NUM_NODES"]))
+        mm_consts["MAX_NUM_ZONES"] = prog["__MAX_NR_ZONES"].value_()
+
+        # Determine page size
+        mm_consts["PAGE_SIZE"] = prog.constant("PAGE_SIZE").value_()
+        mm_consts["PAGE_SHIFT"] = prog.constant("PAGE_SHIFT").value_()
+
+        # Determine architecture-specific parameters
+        if prog.platform.arch == drgn.Architecture.X86_64:
+            # arch/x86/include/asm/pgtable_64_types.h
+            _pmd_shift = 21
+            if "vmalloc_base" in prog:
+                _vmalloc_size_tb = 32
+                mm_consts["VMALLOC_START"] = prog["vmalloc_base"].value_()
+                mm_consts["VMALLOC_END"] = (
+                    mm_consts["VMALLOC_START"] + (_vmalloc_size_tb << 40) - 1
+                )
+            else:
+                mm_consts["VMALLOC_START"] = 0xFFFFC90000000000
+                mm_consts["VMALLOC_END"] = 0xFFFFE8FFFFFFFFFF
+        elif prog.platform.arch == drgn.Architecture.AARCH64:
+            # arch/arm64/include/asm/pgtable-hwdef.h
+            n = 2
+            _pmd_shift = (mm_consts["PAGE_SHIFT"] - 3) * (4 - (n)) + 3
+        else:
+            raise Exception("Target vmcore's architecture is not supported.")
+
+        _hpage_pmd_shift = _pmd_shift
+        _hpage_pmd_order = _hpage_pmd_shift - mm_consts["PAGE_SHIFT"]
+        mm_consts["HPAGE_PMD_NR"] = 1 << _hpage_pmd_order
+
+        # Determine the max number of swap file types.
+        # In mm/swap_state.c, ``nr_swapper_spaces`` is defined as:
+        # "static unsigned int nr_swapper_spaces[MAX_SWAPFILES] __read_mostly;"
+        if "nr_swapper_spaces" in prog:
+            mm_consts["MAX_SWAPFILES"] = len(prog["nr_swapper_spaces"])
+        else:
+            # "nr_swapper_spaces" does not exist in UEK-4.
+            mm_consts["MAX_SWAPFILES"] = len(prog["swapper_spaces"])
+
+        # Save ``mm_consts`` to cache.
+        cache["mm_constants"] = mm_consts
+    return mm_consts
+
+
+def print_val_kb(prog: Program, text: str, num: int) -> None:
+    """
+    Produce the formatted output that matches the output in /proc/meminfo.
+
+    :param prog: drgn program
+    :param text: The output item's name, e.g, "MemTotal"
+    :param num: The output item's value
+    """
+    if num < 0:
+        return
+
+    num_kb = num << (get_mm_constants(prog)["PAGE_SHIFT"] - 10)
+    num_kb_str = str(num_kb)
+    print(f"{text + ':': <15} {num_kb_str: >8} kB")
+
+
+def get_active_numa_nodes(prog: Program) -> List[Object]:
+    """
+    Get a list of all active NUMA nodes.
+
+    :returns: a list of ``struct pglist_data *`` objects.
+    """
+    active_numa_nodes = []
+    node_data = prog["node_data"]
+    for i in range(get_mm_constants(prog)["MAX_NUM_NODES"]):
+        if node_data[i].value_() != 0x0:
+            active_numa_nodes.append(node_data[i])
+    return active_numa_nodes
+
+
+def for_each_node_zone(prog: Program, node: Object) -> Iterator[Object]:
+    """
+    Iterate over all memory zones in a NUMA node.
+
+    :param prog: drgn program
+    :param node: The drgn object of a NUMA node's ``struct pglist_data``.
+    :returns: Iterator of ``struct zone *`` objects.
+    """
+    node_zones = node.node_zones
+    max_num_zones = prog["__MAX_NR_ZONES"].value_()
+    for j in range(max_num_zones):
+        if node_zones[j].value_() != 0x0:
+            yield node_zones[j]
+
+
+def for_each_zone(prog: Program) -> Iterator[Object]:
+    """
+    Iterate over all zones in a system that contains 1+ NUMA nodes.
+
+    :returns: Iterator of ``struct zone *`` objects.
+    """
+    for node in get_active_numa_nodes(prog):
+        for zone in for_each_node_zone(prog, node):
+            yield zone
+
+
+def for_each_hstate(prog: Program) -> Iterator[Object]:
+    """
+    Iterate over all hugepage pools in a system.
+    Note: each |hstate| represents a pool for a certain unit size hugepages of a NUMA node.
+
+    :returns: Iterator of ``struct hstate`` objects.
+    """
+    hstates = prog["hstates"].read_()
+    for i in range(prog["hugetlb_max_hstate"].value_()):
+        yield hstates[i]
+
+
+def get_total_available_pages(prog: Program) -> int:
+    """
+    Get an estimation of the amount of memory available for future use.
+    This includes the number of free memory pages and the number of pages
+    that can be reclaimed from caches.
+
+    :returns: The number of available memory in pages.
+    """
+    global_stats = get_global_mm_stats(prog)
+
+    total_free_pages = global_stats["NR_FREE_PAGES"]
+    total_reserve_pages = prog["totalreserve_pages"].value_()
+    available_pages = total_free_pages - total_reserve_pages
+
+    # Calculate the global low watermark.
+    low_wmark = 0
+    for zone in for_each_zone(prog):
+        if has_member(zone, "_watermark"):
+            zone_low_wmark = zone._watermark[prog.constant("WMARK_LOW")]
+        else:
+            zone_low_wmark = zone.watermark[prog.constant("WMARK_LOW")]
+        low_wmark += zone_low_wmark.value_()
+
+    # Estimate reclaimable page cache.
+    lru_active_file = global_stats["NR_ACTIVE_FILE"]
+    lru_inactive_file = global_stats["NR_INACTIVE_FILE"]
+    pagecache = lru_active_file + lru_inactive_file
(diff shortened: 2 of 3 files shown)


brenns10:
Hey @jianfenw , please double check the internal docs regarding Github profiles. I see that the "oracle-samples" organization membership is public. Did you make sure to add your Oracle email address to your account?

jianfenw:
Thanks, Stephen. The OCA check failed because I didn't make my membership in oracle public when I submitted this PR. I made my memberships in oracle and oracle-samples public afterwards. However, the OCA check didn't get updated automatically. I may have to update the commit to trigger a new check later.

drgn_tools/meminfo.py
```diff
@@ -0,0 +1,755 @@
...
+    # Determine page size
+    _PAGE_SIZE = prog.constant("PAGE_SIZE").value_()
+    _PAGE_SHIFT = prog.constant("PAGE_SHIFT").value_()
+
+    # Set arch
+    if "aarch64" in uts["machine"]:
+        _IS_AARCH_64 = True
+
+        # include/linux/numa.h
+        _NODES_SHIFT = 2
```
brenns10, line 116: So, this is an excellent example of a constant that's really difficult to get. It's defined as a `#define`, and it comes from the kernel config. I'm actually working on a way to make it easy and reliable to access the kernel configuration from Drgn, but this will take a while, so we can't hope for that to work.

But we can still get this without hard-coding it. Notice that below the definition of `NODES_SHIFT`, we get `#define MAX_NUMNODES (1 << NODES_SHIFT)`. And `MAX_NUMNODES` is used in the definition of lots of data structures! I identified one (`preferred_node_policy`) which is defined in `mm/mempolicy.c` as:

```
static struct mempolicy preferred_node_policy[MAX_NUMNODES];
```

This is a pretty unique name and it seems to be present from UEK4 to mainline, so long as `CONFIG_NUMA` is enabled. We can use this to get the desired value:

```python
def nodes_shift(prog: Program) -> int:
    return int(math.log2(len(prog["preferred_node_policy"])))
```

Of course you'd want to include a comment describing why this is done, otherwise it will be incomprehensible to readers.

drgn_tools/meminfo.py
```diff
@@ -0,0 +1,755 @@
...
+        SZ_128M = 0x08000000
+        VA_BITS = 48
+        VA_START = 0xFFFFFFFFFFFFFFFF - (1 << VA_BITS) + 1
+        PAGE_OFFSET = 0xFFFFFFFFFFFFFFFF - (1 << (VA_BITS - 1)) + 1
+        STRUCT_PAGE_MAX_SHIFT = 6
+        KASAN_SHADOW_SIZE = 0
+        MODULES_VADDR = VA_START + KASAN_SHADOW_SIZE
+        MODULES_VSIZE = SZ_128M
+        MODULES_END = MODULES_VADDR + MODULES_VSIZE
+        VMEMMAP_SIZE = 1 << (VA_BITS - _PAGE_SHIFT - 1 + STRUCT_PAGE_MAX_SHIFT)
```
brenns10, line 142: Using the upper case for these variables is confusing. I see that you're using `_` prefix for all your private global variables (that's great!). But since these are locals, the `ALL_CAPS` naming convention is confusing if you're just reading this function.

I know you're making these variables comparable to what you'd find in the corresponding header file, but I'd prefer if you made them lower case so it's easier to read, and also easier to see what is the actual final output of these computations.

drgn_tools/meminfo.py
```diff
@@ -0,0 +1,755 @@
...
+    param num: The output item's value
+    """
+    if num < 0:
+        return
+    text += ":"
+    num_kb = num << (_PAGE_SHIFT - 10)
+    print("%s%s kB" % (text.ljust(16, " "), str(num_kb).rjust(8, " ")))
+
+
+def parse_machine_info(prog: Program) -> None:
```
brenns10, line 87: I'd suggest two approaches for this function - use whichever you like better.

First option is to have it return a dictionary mapping the constant name to its value. You could name the function `get_mm_constants()` or something like that. This can avoid storing everything in a global variable, and you can even use the prog.cache to cache it.

Second option is to simply compute the constants where you need them. Aside from the difficult vmalloc constants, I think most of them are easy to just compute whenever you want them.

drgn_tools/meminfo.py
```diff
@@ -0,0 +1,755 @@
...
+    try:
+        active_numa_nodes = prog.cache["active_numa_nodes"]
+    except KeyError:
+        active_numa_nodes = []
+        node_data = prog["node_data"]
+        for i in range(_MAX_NUM_NODES):
+            if node_data[i].value_() != 0x0:
+                active_numa_nodes.append(node_data[i])
+        prog.cache["active_numa_nodes"] = active_numa_nodes
+    return active_numa_nodes
```
brenns10, line 185: Please store everything for this module within a cache dictionary named `drgn_tools.meminfo`. EG:
```suggestion
    try:
        cache = prog.cache.setdefault("drgn_tools.meminfo", {})
        active_numa_nodes = cache["active_numa_nodes"]
    except KeyError:
        active_numa_nodes = []
        node_data = prog["node_data"]
        for i in range(_MAX_NUM_NODES):
            if node_data[i].value_() != 0x0:
                active_numa_nodes.append(node_data[i])
        cache["active_numa_nodes"] = active_numa_nodes
    return active_numa_nodes
```

drgn_tools/meminfo.py
```diff
@@ -0,0 +1,755 @@
...
+
+
+def for_each_node_zone(node: Object) -> Iterator[Object]:
+    """
+    Iterate over all memory zones in a NUMA node.
+
+    :returns: Iterator of ``struct zone *`` objects.
```
brenns10, line 192: Please document the `node` parameter's type.

drgn_tools/meminfo.py
```diff
@@ -0,0 +1,755 @@
...
+    """
+    Iterate over all block devices in a system.
+
+    :returns: Iterator of ``struct inode *`` objects.
+    """
+    return list_for_each_entry(
+        "struct inode",
+        prog["blockdev_superblock"].member_("s_inodes").address_of_(),
+        "i_sb_list",
+    )
```
brenns10, line 232: My first thought was that these should go into `drgn_tools.block`.

But I now see that they pretty much exist for the purpose of `get_blockdev_pages()`. I may be wrong, but I don't think these have much of a use case beyond that function? I think it would be fine if you were to move these into `get_blockdev_pages()`

drgn_tools/meminfo.py
```diff
@@ -0,0 +1,755 @@
...
+    :returns: Iterator of ``struct hstate`` objects.
+    """
+    hstates = prog["hstates"].read_()
+    for i in range(prog["hugetlb_max_hstate"].value_()):
+        yield hstates[i]
+
+
+def is_valid_node_page_state(prog: Program, key: str, offset: int = 0) -> bool:
+    try:
+        valid_node_states = prog.cache["valid_node_states"]
```
brenns10, line 249: As above, put things into a cache object named `drgn_tools.meminfo`.

drgn_tools/meminfo.py
```diff
@@ -0,0 +1,755 @@
...
+def get_lru_pages(prog: Program) -> List[int]:
+    """
+    Get the number of pages for all LRU lists, including: INACTIVE_ANON,
+    ACTIVE_ANON, INACTIVE_FILE, ACTIVE_FILE, and LRU_UNEVICTABLE.
+
+    :returns: A list of integer numbers that represents numbers of pages
+    in each LRU list.
+    """
+    try:
+        lru_pages = prog.cache["lru_pages"]
```
brenns10, line 440: Same as the above with the caches

drgn_tools/meminfo.py
```diff
@@ -0,0 +1,755 @@
...
+    except KeyError:
+        num_lru_lists = prog.constant("NR_LRU_LISTS").value_()
+        lru_pages = [0 for _ in range(num_lru_lists)]
+        for lru in range(num_lru_lists):
+            lru_pages[lru] = get_global_page_state(prog, "NR_LRU_BASE", lru)
```
brenns10, line 445: I would argue that rather than doing a list comprehension, followed by a loop, you could do this in one comprehension.
```suggestion
        lru_pages = [
            get_global_page_state(prog, "NR_LRU_BASE", lru)
            for lru in range(num_lru_lists)
        ]
```
It's not any shorter but I think it makes a bit more sense on reading.

drgn_tools/meminfo.py
```diff
@@ -0,0 +1,755 @@
...
+
+    direct_4k = direct_pages_count[prog.constant("PG_LEVEL_4K")].value_()
+    print("DirectMap4k:    %8d kB" % (direct_4k << 2))
+
+    if _IS_X86_64:
+        direct_2m = direct_pages_count[prog.constant("PG_LEVEL_2M")].value_()
+        print("DirectMap2M:    %8d kB" % (direct_2m << 11))
+    else:
+        direct_4m = direct_pages_count[prog.constant("PG_LEVEL_2M")].value_()
+        print("DirectMap4M:    %8d kB" % (direct_4m << 12))
```
brenns10, line 534: Is this for supporting x86 32 bit?

I wouldn't worry about that :)

drgn_tools/util.py
```diff
@@ -123,6 +123,23 @@ def enum_name_get(
...
+    :param offset: The target enum item's offset
+    :returns: The target enum item's index
+    """
+    try:
+        enum_item = prog.constant(key)
+    except LookupError:
+        raise IndexError("UNKNOWN enum %s" % (key))
+    return enum_item.value_() + offset
+
+
```
brenns10, line 142: I'm having trouble understanding the purpose of this function. Why use it instead of `int(prog["key"]) + offset`?

It seems like most users don't use the offset anyway.

brenns10 (changes requested):
This is a pretty excellent job! You managed to fit a lot of logic into this. Most of my review comments have to do with the use of global variables, but the logic itself seems sound.

doc/api.rst
```diff
@@ -106,9 +106,14 @@ drgn_tools.rds
 .. automodule:: drgn_tools.rds
    :members:
 
-
```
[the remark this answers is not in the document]
jianfenw: Added lines back

drgn_tools/meminfo.py
jianfenw: Sounds good. Updated.

drgn_tools/meminfo.py
jianfenw: Done

drgn_tools/meminfo.py
jianfenw: Makes sense. Thanks. Updated.

drgn_tools/meminfo.py
jianfenw: Done

drgn_tools/meminfo.py
jianfenw: Ah, this is for non x86-64. I was not aware of that meminfo won't include DirectMap statistics for aarch64 when I implemented this function. Later, I added the above ``_IS_AARCH_64`` check and forgot to delete this part.
Deleted.

drgn_tools/util.py
jianfenw: Yes, it is a unified function that checks that the |key| exists in |prog| and also provides the 'offset' functionality. It seems that the offset part is only used in calculating LRU lists' statistics.
I think it makes sense to just calculate the enum value inline. Updated.

drgn_tools/meminfo.py
jianfenw: Yes. These two functions are only used by ```get_blockdev_pages()``` for now.
I don't anticipate any special use case of iterating all block devices' inodes. I am okay to just keep them inline because they are pretty short helper functions. If we do need them later, we can move this to ```drgn_tools.block```.

drgn_tools/meminfo.py
jianfenw: Added

drgn_tools/meminfo.py
```diff
@@ -0,0 +1,755 @@
...
+# Handling of hardware poisoned pages with memory corruption.
+_SWP_HWPOISON_NUM = 1
+
+# |MAX_SWAPFILES| defines the maximum number of swaptypes.
+_MAX_SWAPFILES = (
+    (1 << _MAX_SWAPFILES_SHIFT)
+    - _SWP_DEVICE_NUM
+    - _SWP_MIGRATION_NUM
+    - _SWP_HWPOISON_NUM
+)
```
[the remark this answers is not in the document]
jianfenw: Thanks for these great advices. Let me combine these rules of thumb when addressing your detailed comments below.

drgn_tools/meminfo.py
jianfenw: Thanks for the hint. I've made changes to these constant variables.

drgn_tools/meminfo.py
jianfenw: Updated accordingly.

drgn_tools/meminfo.py
jianfenw: I adopted the first option because these variables are used in many functions. Caching them can save some memory lookups.

drgn_tools/meminfo.py
```diff
@@ -0,0 +1,755 @@
...
+    global _VMALLOC_END, _VMALLOC_START, _VMALLOC_SPACE
+
+    uts = get_uts(prog)
+    # Set UEK version
+    if "4.14.35" in uts["release"]:
+        _UEK_VERSION = 5
+    elif "5.4.17" in uts["release"]:
+        _UEK_VERSION = 6
+    else:
+        _UEK_VERSION = 7
```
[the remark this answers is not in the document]
jianfenw: Agreed on your comment. This is helpful.
I deleted the ```_UEK_VERSION``` variable and do not use any UEK version in the code logic.
For the ```for_each_hstate()``` case, it is actually not about whether the ```hstates``` variable exists or not. It is just that kernels of some UEK versions simply do not output the information. I think we can just remove that specific check.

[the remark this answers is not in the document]
jianfenw: Also, added the commit ID for some other places in the comment to improve the code readability.

drgn_tools/meminfo.py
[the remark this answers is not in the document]
jianfenw: Done.

jianfenw:
I've resolved comments and also included uek4 support. This branch has passed all CI tests.
Please review again. Thank you.

drgn_tools/meminfo.py
```diff
@@ -0,0 +1,767 @@
...
+            if item == "HardwareCorrupted":
+                print("HardwareCorrupted: %5d kB" % (stats[item]))
+            else:
+                print(
+                    "%s%s kB"
+                    % (
+                        (item + ":").ljust(16, " "),
+                        str(stats[item]).rjust(8, " "),
+                    )
+                )
```
brenns10, line 747: I'm not certain I understand the need for the two cases? Also, I think you could shorten this code quite a bit by using f-strings. The following might not be _exactly_ equivalent to your code, but it's a similar idea:

```python
print(f"{item + ':':<15} {stats[item]: 8d} kB")
```

brenns10 (changes requested):
This is looking good, but still needs some work around caching. This needs to work correctly on live systems, and give always up-to-date statistics each time you run it. That means you need to limit the use of caching. I've provided several suggestions inline.

drgn_tools/meminfo.py
```diff
@@ -0,0 +1,767 @@
...
+        cache = prog.cache.setdefault("drgn_tools.meminfo", {})
+        lru_pages = cache["lru_pages"]
+    except KeyError:
+        num_lru_lists = prog.constant("NR_LRU_LISTS").value_()
+        lru_pages = [
+            get_global_page_state(prog, "NR_LRU_BASE", lru)
+            for lru in range(num_lru_lists)
+        ]
+        cache["lru_pages"] = lru_pages
+    return lru_pages
```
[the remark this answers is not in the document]
jianfenw: Thanks. This definitely clarifies cases of using caching in drgn modules.
The three guidelines make sense to me. For memory-related module, I expect that many variables, especially statistics, are constantly changing on live vmcores. I will update my commit accordingly.

drgn_tools/meminfo.py
```diff
@@ -0,0 +1,767 @@
...
+    if key == "NR_ANON_MAPPED" and key not in prog:
+        key = "NR_ANON_PAGES"
+    if key == "NR_KERNEL_STACK_KB" and key not in prog:
+        key = "NR_KERNEL_STACK"
+
+    if is_valid_zone_page_state(prog, key):
+        return get_global_zone_page_state(prog, key, offset)
+    if is_valid_node_page_state(prog, key):
+        return get_global_node_page_state(prog, key, offset)
+    raise IndexError("INVALID enum item %s in node/zone statistics", key)
```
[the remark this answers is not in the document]
jianfenw: This is definitely amazing. Thanks for your suggestion.

When I implemented this module, I basically followed the way of reading global statistics counters in the kernel: by having functions like: global_node_page_state(xxx) and global_zone_page_state(xxx). Then, I realized that I might be reading ```vm_node_stat``` and ```vm_zone_stat``` many many times, so I decided to cache them.

I agree that we should avoid caching because we'd like to support live vmcores. I like the way done in your suggestion in that we can parse and store all statistics items all at once without further caching them.

drgn_tools/meminfo.py
jianfenw: This is to keep the original spacing as done in the kernel implementation.
"HardwareCorrupted:" has a length 18 and exceeds the target text 15 by 3, so the statistics number's length is reduced from 8 to 5.

I added a comment to explain this and also unified methods of formatting strings in this meminfo module.

drgn_tools/meminfo.py
```diff
@@ -0,0 +1,767 @@
...
+        # "static unsigned int nr_swapper_spaces[MAX_SWAPFILES] __read_mostly;"
+        # "struct address_space *swapper_spaces[MAX_SWAPFILES] __read_mostly;"
+        if "nr_swapper_spaces" in prog:
+            mm_consts["MAX_SWAPFILES"] = len(prog["nr_swapper_spaces"])
+        else:
+            # "nr_swapper_spaces" does not exist in UEK-4.
+            mm_consts["MAX_SWAPFILES"] = len(prog["swapper_spaces"])
+
+        cache = mm_consts
+    return mm_consts
```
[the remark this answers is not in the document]
jianfenw: Updated accordingly. Thanks for the catch.

drgn_tools/meminfo.py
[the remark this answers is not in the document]
jianfenw: Done.

jianfenw:
Thank you Stephen. I have updated the commit according to your comments. Appreciate all your wonderful suggestions.

drgn_tools/meminfo.py
```diff
@@ -0,0 +1,678 @@
...
+            _modules_vsize = _sz_128m
+            _modules_end = _modules_vaddr + _modules_vsize
+            _vmemmap_size = 1 << (
+                _va_bits - mm_consts["PAGE_SHIFT"] - 1 + _struct_page_max_shift
+            )
+
+            mm_consts["VMALLOC_START"] = _modules_end
+            mm_consts["VMALLOC_END"] = (
+                _page_offset - _pud_size - _vmemmap_size - _sz_64k
+            )
```
brenns10, line 159: `VMALLOC_START` can't be reliably computed on aarch64. See commit torvalds/linux@9ad7c6d5e75b. I don't really have a suggestion of how to handle it though. You can see how I handle this in `drgn_tools/mm.py`. The computation here overlaps to a decent extent with that. You could make `AddrKind._ranges` public and try to reuse the value I compute (but there's no guarantee it's fully correct, just "good enough").

Or, you could simply omit it, and don't provide the VmallocTotal field on aarch64, which seems fine to me.

drgn_tools/meminfo.py
```diff
@@ -0,0 +1,678 @@
...
+        """
+        if key in self.renames and key not in self.data:
+            key = self.renames[key]
+        return super().__getitem__(key)
+
+
+def _read_stats(prog: Program, arr_name: str, enum: str) -> UserDict:
+    """
+    Read counters from a global statistics array with name ```arr_name```.
+    Parse all statistics items that are indexed by ```enum``` and store results
```
brenns10, line 50: nit: use two backticks for the inline code:
```suggestion
    Read counters from a global statistics array with name ``arr_name``.
    Parse all statistics items that are indexed by ``enum`` and store results
```
this applies everywhere and should be a quick fix. It's restructured text format, in case you're wondering.

drgn_tools/meminfo.py
```diff
@@ -0,0 +1,678 @@
...
+    total_reserve_pages = prog["totalreserve_pages"].value_()
+    available_pages = total_free_pages - total_reserve_pages
+
+    # Calculate the global low watermark.
+    low_wmark = 0
+    for zone in for_each_zone(prog):
+        if has_member(zone, "_watermark"):
+            zone_low_wmark = zone.member_("_watermark")[
+                prog.constant("WMARK_LOW")
+            ]
```
brenns10, line 269: nit: (which applies throughout the PR) - the `member_()` method is almost always unnecessary. You could just write `zone._watermark[prog.constant("WMARK_LOW")]` here. The [`member_()`](https://drgn.readthedocs.io/en/latest/api_reference.html#drgn.Object.member_) method exists mainly to aid you in instances where a method on `Object` has the same name as a struct field. For example, if there was a field named `value_` in a struct, then you'd need to use `obj.member_("value_")` since `value_` is already a method of drgn's Object.

That said, there's nothing technically wrong with using it! I guess I'm just pointing it out to let you know, and it is slightly easier to read without it.

brenns10 (commented):
I have just three little nits here, overall this is good. The most difficult one here is with respect to `VMALLOC_END` on aarch64, and I'd honestly suggest you just omit the vmalloc info on aarch64, and omit the vmalloc statistic when you go to print everything out.

I checked the internal CI and the failures all have to do with the currently quite flaky in-flight I/O test. I may need to disable that test or find some way to make it a bit better. In any case, all the vmcores passed.

Overall it's looking good and I expect to merge it after these quick changes. Thank you!

drgn_tools/meminfo.py
jianfenw: Sounds great. Thanks for pointing it out.

drgn_tools/meminfo.py
jianfenw: Done!

drgn_tools/meminfo.py
jianfenw: I kept it calculated this way because this field has been checked and passed all vmcore tests, which means the parsed value matched exactly with the output from /proc/meminfo.

With your input, I agree that the vm layout may be changed by kernel commits. Plus, since you've already had a function to calculate address regions for vms, I'd just omit the field for aarch64.

jianfenw:
Updated this commit accordingly! Thank you, Stephen.

brenns10 (approved):
Thanks for iterating on this, it looks great! I checked the tests, and the only failures on the vmcores were the ones on the in-flight I/O test, which are fixed in #28. So it's good to merge!

PR #6, opened 2023-10-28
Add aarch64 support for cpuinfo module

Hannibal404:


Hannibal404:
Fixes #8 with [823f9cb](https://github.com/oracle-samples/drgn-tools/pull/6/commits/823f9cb3b393f7b2b831b1a6f6b2b859ac3efd8c)

tests/test_cpuinfo.py
```diff
@@ -8,40 +9,76 @@
 def test_cpuinfo(prog):
     cpuinfo.print_cpu_info(prog)
 
-    if not (ProgramFlags.IS_LIVE & prog.flags) or not (
-        prog["init_uts_ns"].name.machine.string_().decode("utf-8") == "x86_64"
-    ):
+    if not (ProgramFlags.IS_LIVE & prog.flags):
         return
```
brenns10, line 13: Let's go ahead and use `pytest.skip()` here. This is just a bit more honest about what's happening in the test - we can't test it so we're skipping the rest of it. Rather than returning, which indicates "we were able to test it, and it passed!"

By doing this after we've run `print_cpu_info()`, we still get the "smoke test" out of it -- if an error with the types, etc, crops up, the test would fail before it gets skipped.

```suggestion
        pytest.skip("Supported only on live systems")
```

You'll need an import at the top as well:
```python
import pytest
```

[Documentation](https://docs.pytest.org/en/7.0.x/how-to/skipping.html)

brenns10 (changes requested):
Just one tiny nit regarding tests, but otherwise it looks good.

I rebased it on top of #20 and ran the vmcore tests against the full suite - it passes, and fixes the one regression in #8. Thanks for this!

tests/test_cpuinfo.py
brenns10: Actually, this is fine as is. I'm going to merge it now :)

brenns10 (approved):

PR #7, opened 2023-10-30
Improvements for CI Testing

brenns10:
This pull request incorporates several improvements for CI testing:

1. The tox environment names I established for Github Actions (`py3.6`, ...) don't seem to be correct. They need to have names like `py36` in order to work correctly. So, the first commit changes the names of these environments and makes the necessary changes for the Github Actions workflow to use those names instead. This makes Github Actions choose the correct Python version, and it also improves the very slow CI test runtimes we've seen on Gitlab CI.
2. Test output is very cluttered, and the test runner for Gitlab had the ability to split it into collapsible sections. I've added support to do this for Github too, and now the litevm test runner uses collapsible sections to help organize output.
3. I've been going through the developer workflow documentation to ensure everything works, so there are a couple tweaks to the Makefile and the heavyvm testing to make everything work correctly. There are actually quite a few updates to the heavyvm testing which simplify the dependencies and make sure everything works on OL9.

### DIFF
--- a/.github/workflows/litevm.yml
+++ b/.github/workflows/litevm.yml
@@ -24,10 +24,14 @@ jobs:
     strategy:
       matrix:
         # OL7 and OL8 use 3.6. OL9 uses 3.9. Also test on the latest.
-        python-version: ["3.6", "3.9", "3.12"]
+        python-minor: ["6", "9", "12"]
       fail-fast: false
     env:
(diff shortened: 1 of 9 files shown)


brenns10:
All CI tests passing, internal and external!

brenns10:
You can see one of the changes in action in the CI:
image
Test runs are now collapsed into sections for each UEK, which will make it easier to browse results.

biger410 (approved):
These patches looks good. Thanks

PR #16, opened 2023-11-03
nfs_tools: Helpers for NFS (client-side only) and SUNRPC subsystem.

daimngo:
Drgn helpers to display NFS and SUNRPC major data structures in vmcore or live system.

This initial version has support for NFS client-side only. NFS server-side support will be added in future enhancement.

### DIFF
--- a/drgn_tools/nfs_tools.py
+++ b/drgn_tools/nfs_tools.py
@@ -1,0 +1,865 @@
+# Copyright (c) 2023, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
+"""
+Helpers for NFS (client-side only) and SUNRPC subsystem.
+"""
+import argparse
+import errno
+import ipaddress
+import socket
+import typing as t
+from itertools import chain
+
+import drgn
+from drgn import Object
+from drgn import Program
+from drgn.helpers.common.format import escape_ascii_string
+from drgn.helpers.common.type import enum_type_to_class
+from drgn.helpers.linux.fs import d_path
+from drgn.helpers.linux.fs import for_each_mount
+from drgn.helpers.linux.fs import inode_path
+from drgn.helpers.linux.list import list_count_nodes
+from drgn.helpers.linux.list import list_for_each_entry
+
+from drgn_tools.corelens import CorelensModule
+from drgn_tools.itertools import take
+from drgn_tools.rds import rds_inet_ntoa
+from drgn_tools.util import BitNumberFlags
+from drgn_tools.util import enum_name_get
+from drgn_tools.util import has_member
+from drgn_tools.util import redirect_stdout
+
+__all__ = (
+    "nfsshow",
+    "show_nfs_clients",
+    "show_rpc_clnts",
+    "show_rpc_xprts",
+    "show_rpc_tasks",
+)
+
+""" include/linux/sunrpc/sched.h """
+
+
+class RpcTaskState(BitNumberFlags):
+    RPC_TASK_RUNNING = 0
+    RPC_TASK_QUEUED = 1
+    RPC_TASK_ACTIVE = 2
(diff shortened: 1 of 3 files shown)


drgn_tools/nfs_tools.py
```diff
@@ -0,0 +1,931 @@
...
+
+def panic_task(prog: drgn.Program) -> Object:
+    """
+    Return the task_struct that causes the panic.
+
+    :param prog: drgn program
+    :returns: task_struct object
+    """
+
+    return per_cpu(prog["runqueues"], prog["crashing_cpu"]).curr
```
brenns10, line 915: You can use `prog.crashed_thread()` to get the `drgn.Thread` which is crashing. `Thread.object` will be a `struct task_struct *`.

Even aside from that, it's important to note that `prog["crashing_cpu"]` is x86_64 specific, and we support aarch64, so if you were implementing this, I would recommend `prog["panic_cpu"]`.

drgn_tools/nfs_tools.py
```diff
@@ -0,0 +1,931 @@
...
+
+    :param prog: drgn program
+    :param struct_name: name of kernel structure
+    :param value: kernel address
+    :returns: Object of the struct
+    """
+
+    name = "struct " + struct_name + " *"
+    obj = Object(prog, name, value=addr)
+    return obj
```
brenns10, line 931: I don't see this being  used anywhere, do we need to include it?

drgn_tools/nfs_tools.py
```diff
@@ -0,0 +1,931 @@
...
+        )
+        kernel_version = vers[0]
+        uek5 = kernel_version.split(".")[0] == "4"
+    except ValueError:
+        print(
+            "WARNING: don't recognize kernel version %s"
+            % prog["UTS_RELEASE"].string_().decode("ascii")
+        )
+        uek5 = False
+    return uek5
```
brenns10, line 802: I think we can delete this, I've made notes at all the call sites.

drgn_tools/nfs_tools.py
```diff
@@ -0,0 +1,931 @@
...
+
+    clients = nfs_client_list(prog)
+    print(
+        "============================== NFS_CLIENT ================================"
+    )
+    for clnt in clients:
+        show_nfs_client(prog, clnt)
+
+
+def nfsshow(prog: drgn.Program, max_tasks: int = 10) -> None:
```
brenns10, line 770: It looks like this is the main entry point where you'd expect most people to call for summary NFS info. I'd recommend that you create a corelens module to call this function. It would look something like this:
```python
class NfsShow(CorelensModule):
    """Print summary information on NFS client side"""

    name = "nfsshow"

    # there's no point in running unless the nfs module is loaded
    skip_unless_have_kmod = "nfs"

    # we need sunrpc, and all nfs related debuginfo loaded
    debuginfo_kmods = ["sunrpc", "nfs", "nfs*"]

        parser.add_argument(
            "--max-tasks",
            type=int,
            default=10,
            help="list at most <number> rpc_tasks",
        )

    def run(self, prog: Program, args: argparse.Namespace) -> None:
        nfsshow(prog, max_tasks=args.max_tasks)
```

brenns10 (changes requested):
This is quite well-organized and well written code. Most of my comments are just getting the coding style and best practices aligned with what drgn-tools expects.

The thing which I didn't cover yet in this review was the testing. I'll follow-up via email about testing. The overview is that we'll have you generate a vmcore from a test system (ideally one each for UEK4-7), and we'll include those into our testing library. I'll be out much of the next two weeks for LPC and Thanksgiving though, so please bear with me on the timeline!

brenns10:
@daimngo I've been out for conference / holiday stuff for the past few weeks. Please do let me know if you're still working on this branch, or if it is ready for another round of review. Thanks!

brenns10 (approved):

PR #22, opened 2023-11-07
block: Reduce the chance of inconsistency for inflight I/O

brenns10:
When reading I/O requests out of a live kernel, there's a pretty good chance that the I/O could complete and the data structure could get freed during the time between when we found the data structure, and when we actually print it.

This means that the list iterators, and the printing functions, all risk reading stale and potentially overwritten data.

We can help reduce the chances of problems with two approaches:

1. Read the requests into a list ahead of time. This will help avoid the list iterators getting stale data.

2. Once we have all the requests, dereference the pointer and use .read_() to get a copy of the data that won't change even as the kernel memory changes.

With these two changes, the chance of reading false data is reduced (but not eliminated). It should reduce some flakiness in tests.

### DIFF
--- a/drgn_tools/block.py
+++ b/drgn_tools/block.py
@@ -196,7 +196,7 @@ def rq_pending_time_ms(rq: Object) -> int:
     """
     Get io pending time in ms
 
-    :param rq: ``struct request *``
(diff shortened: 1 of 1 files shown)


biger410 (approved):
Looks good to me

PR #23, opened 2023-11-07
corelens: automatically import all modules

brenns10:
With so much parallel development going on in drgn-tools, we're hitting a lot of conflicts around the list of modules which contain corelens modules.

I had wanted to avoid importing all of drgn_tools when I initially wrote this, since it seemed expensive and unnecessary given that most of the helpers didn't have corelens modules.

Now, it makes much more sense, since corelens modules are being written all the time. So let's avoid the merge conflicts, and just automatically import everything.

### DIFF
--- a/drgn_tools/corelens.py
+++ b/drgn_tools/corelens.py
@@ -9,6 +9,7 @@ import collections
 import importlib
 import inspect
 import os
+import pkgutil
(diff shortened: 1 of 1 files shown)


biger410 (approved):
Looks good to me.

PR #25, opened 2023-11-08
Add sysinfo helper

srd-17:


drgn_tools/sys.py
```diff
@@ -0,0 +1,145 @@
...
+
+from drgn_tools.corelens import CorelensModule
+from drgn_tools.cpuinfo import aarch64_get_cpu_info
+from drgn_tools.cpuinfo import x86_get_cpu_info
+from drgn_tools.mm import totalram_pages
+from drgn_tools.table import print_dictionary
+from drgn_tools.util import human_bytes
+
+
+def load_average(prog: Program) -> str:
```
brenns10, line 26: Given that `loadavg` already exists, and the function of this is to simply turn it into a string that looks like what you want, a better name would be `loadavg_str`

drgn_tools/sys.py
```diff
@@ -0,0 +1,145 @@
...
+                task_state = task.state
+        except LookupError:
+            task_state = task.__state
+        if task_state == 0:
+            task_running += 1
+        elif task_state == 1:
+            task_int += 1
+        elif task_state == 2:
+            task_unint += 1
+    return f"{int(tasks_count)} [R:{task_running} D:{task_unint} S:{task_int}]"
```
brenns10, line 70: * Drgn conveniently has a helper to convert the states to character codes: [`task_state_to_char()`](https://drgn.readthedocs.io/en/latest/helpers.html#drgn.helpers.linux.sched.task_state_to_char). You can use this so that you don't need to hard-code anything.
* You seem to have over-complicated your `task.__state` logic, you can use an if statement, or a try/except, but you don't need both. But in any case, with the above helper, you don't need that either.
* I'm a bit confused why `tasks_count` starts at `prog["nr_cpu_ids"]` and not 0?
```suggestion
    state_count = {}
    tasks_count = 0
    for task in for_each_task(prog):
        tasks_count += 1
        state_code = task_state_to_char(task)
        state_count[state_code] = state_count.get(state_code, 0) + 1
    # ensure the first categories are R(unning), D (uninterruptible), S (interruptible)
    output = f"{tasks_count}"
    for code in ("R", "D", "S"):
        output += " {code}:{state_count.get(code, 0)}"
        state_count.pop(code)
    # the remaining items in any order
    for code, count in state_count.items():
        output += " {code}:{count}"
    return output
```

The benefit here being that you get other state codes too: e.g. Z, I, X

brenns10 (changes requested):
Looks good, just a few nits here.

brenns10:
One additional request: I see you used a merge commit to merge `main` into your branch. When updating your branch, please use `git rebase` so we don't get merge commits, and we can have a clean, linear git history.

brenns10 (approved):

PR #26, opened 2023-11-09
Corelens module: dentrycache

brenns10:
We already have some excellent dentry helpers, but no actual corelens module. This PR adds the corelens module, and it adds a bit of optimization so that the runtime is pretty decent.

1. Add the corelens module.
2. Create a `dentry_path_any_mount()` helper to ensure we get a full path to the root for each dentry.
3. Stop loading dentries into memory when we're printing them out, this ensures we can scale to millions of dentries without any issue.
4. Try to match the `oled dentrycache` output by default.
5.  Improve performance by iterating over the dentry hash table in chunks, rather than one entry at a time.

### DIFF
--- a/drgn_tools/dentry.py
+++ b/drgn_tools/dentry.py
@@ -294,23 +294,30 @@ def d_count(dentry: Object) -> int:
     return dentry.d_lockref.count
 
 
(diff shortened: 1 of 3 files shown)


biger410 (approved):
LGTM

PR #27, opened 2023-11-15
Add bt module to print all stack traces

Hannibal404:


drgn_tools/bt.py
```diff
@@ -441,3 +444,22 @@ def bt_has(
...
 
     return frame_list
+
+
+def print_all_bt(prog: Program) -> None:
+    """
+    Prints the stack trace of all tasks
+    """
+    for task in for_each_task(prog):
+        bt(prog, pid=task.pid.value_())
```
brenns10, line 454: * I think the way to do this, which would be much nicer for everyone to read, would be to first print all on-CPU tasks, and then print all the remaining tasks. Give each a header so that it's easy to see where the on-CPU tasks begin, and the off-CPU tasks continue.
* I believe that you can just use `bt(task)` instead :)

brenns10 (changes requested):
One change I'd like to see, otherwise this is good, thanks!

brenns10 (approved):

PR #28, opened 2023-11-22
block: save hwq address before coverting it to value obj

biger410:
Has to get hwq address before it was converted to value object, otherwise the dump will fail


brenns10 (approved):
Good catch, thank you.

PR #30, opened 2023-11-24
task: add "ps_m" corelens module

SiddhiK29:


drgn_tools/task.py
```diff
@@ -161,3 +169,93 @@ def task_cpu(task: drgn.Object) -> int:
...
+
+class Taskinfo(CorelensModule):
+    """
+    Corelens Module for ps -m
+    """
+
+    name = "ps_m"
+
+    def run(self, prog: drgn.Program, args: argparse.Namespace) -> None:
+        show_ps_m_info(prog)
```
brenns10, line 261: The end goal is not to have a `ps_m` helper, but have a helper named `ps` that you can run as `ps -m`. To that end, we should rename this to `ps` and use the argument parsing. This suggestion gives the general idea:
```suggestion
class Taskinfo(CorelensModule):
    """
    Corelens Module for ps -m
    """

    name = "ps"
    
    def add_arguments(self, parser: argparse.ArgumentParser) -> None:
        parser.add_argument("-m", dest="last_run", action="store_true", help="show last run information")

    def run(self, prog: drgn.Program, args: argparse.Namespace) -> None:
        if args.last_run:
            show_ps_m_info(prog)
        else:
            raise NotImplementedError("currently, only ps -m is implemented")
```

drgn_tools/task.py
```diff
@@ -161,3 +169,93 @@ def task_cpu(task: drgn.Object) -> int:
...
+
+
+def get_command(task: Object) -> str:
+    """
+    :returns: name of the command
+    """
+    return task.comm.string_().decode("utf-8")
+
+
+def show_ps_m_info(prog: Program) -> None:
```
brenns10, line 213: Assuming you're going to eventually expand this, so that it works as just a part of the "ps" command, then you probably want this to accept an iterable of task objects, and simply print them. Further, I don't think "ps_m" is at all a descriptive name, maybe `show_tasks_last_runtime()`?
```suggestion
def show_tasks_last_runtime(tasks: Iterable[Object]) -> None:
```

Then down below, rather than doing `for t in for_each_task(prog)`, you would do `for t in tasks`. This will allow you to implement the filtering logic separately (e.g. for user / kthread / group leader options `-k|-u|-g`).

drgn_tools/task.py
```diff
@@ -161,3 +169,93 @@ def task_cpu(task: drgn.Object) -> int:
...
+     Z    zombie
+    """
+    return task_state_to_char(task)
+
+
+def get_command(task: Object) -> str:
+    """
+    :returns: name of the command
+    """
+    return task.comm.string_().decode("utf-8")
```
brenns10, line 210: This is a case where I would encourage you to use `escape_ascii_string()` (rather than `.decode("utf-8")`). The task command is fully controlled by userspace, and I don't know if we really have any guarantee that it's going to use any particular encoding.

drgn_tools/task.py
```diff
@@ -161,3 +169,93 @@ def task_cpu(task: drgn.Object) -> int:
...
+     R    running or runnable
+     S    interruptible sleep
+     T    stopped by job control signal
+     t    stopped by debugger during the tracing
+     X    dead (should never be seen)
+     Z    zombie
+    """
+    return task_state_to_char(task)
+
+
```
brenns10, line 205: Why does this function exist, if it simply calls the drgn helper? Please delete it, unless there's a reason I'm not seeing.

drgn_tools/task.py
```diff
@@ -161,3 +169,93 @@ def task_cpu(task: drgn.Object) -> int:
     if has_member(task, "cpu"):
         return task.cpu.value_()
     return task_thread_info(task).cpu.value_()
+
+
+def translate_nanoseconds(task: Object) -> str:
```
brenns10, line 174: This function is misnamed. It does not translate nanoseconds. It returns a string representation of the time since the task was last run. Could this be `task_lastrun2now_str()`?

Alternatively, you could instead have the function accept a number of nanoseconds, and convert it into a string representation of the time, and name it something like `format_nanosecond_duration()`

drgn_tools/task.py
```diff
@@ -161,3 +169,93 @@ def task_cpu(task: drgn.Object) -> int:
...
+    """
+    :returns: conversion of nanoseconds to [dd hh:mm:ss.ms] format
+    """
+    nanosecs = task_lastrun2now(task)
+    secs = nanosecs_to_secs(nanosecs)
+    time = str(datetime.timedelta(seconds=secs))
+    return time
```
brenns10, line 181: This returns a string like `83 days, 19:15:12.102000` whereas the crash command outputs in format `0 00:00:00.000`. While I'm not normally a stickler for _exact_ compatibility, it seems that it's important to me to output things roughly the same, since the table alignment would look strange otherwise. You'll just have to format the string yourself to get that.

drgn_tools/task.py
```diff
@@ -161,3 +169,93 @@ def task_cpu(task: drgn.Object) -> int:
...
+    for t in for_each_task(prog):
+        cpu = task_cpu(t)
+        pid = get_pid(t)
+        state = get_ps_state(t)
+        task = find_task(prog, pid)
+        command = get_command(t)
+        last_arrival = translate_nanoseconds(t)
+        nanosecs = task_lastrun2now(t)
+        tasks.append(
+            [
```
brenns10, line 235: Nit: make `tasks` a list of tuples, rather than a list of lists:
```suggestion
            (
```

brenns10 (changes requested):

brenns10:
Please also include tests for these functions, at least smoke tests.

drgn_tools/task.py
```diff
@@ -161,3 +170,91 @@ def task_cpu(task: drgn.Object) -> int:
...
+
+def format_nanosecond_duration(nanosecs: int) -> str:
+    """
+    :returns: conversion of nanoseconds to [dd hh:mm:ss.ms] format
+    """
+    secs = nanosecs_to_secs(nanosecs)
+    dd, rem = divmod(secs, 86400)
+    hh, rem = divmod(rem, 3600)
+    mm, secs = divmod(rem, 60)
+    return "%02ld %02ld:%02ld:%.3f" % (dd, hh, mm, secs)
```
brenns10, line 183: For the seconds component, you should use `06.3f` as the format argument.
- 0 is the padding character
- 3 is the precision, the number of digits after the decimal place to show
- 6 is the total width of the field, including the `.` - using 6 ensures that you always get `xx.xxx`
```suggestion
    return "%02ld %02ld:%02ld:%06.3f" % (dd, hh, mm, secs)
```

drgn_tools/task.py
```diff
@@ -161,3 +170,91 @@ def task_cpu(task: drgn.Object) -> int:
...
+                    pid,
+                    hex(task.value_()),
+                    cpu,
+                    command,
+                ),
+            )
+        )
+    tasklist.sort()
+    for t in tasklist:
+        print(t[1])
```
brenns10, line 238: I really should have noticed this in the first round of review, but we have a function `print_table()` which is really great. I looked at the output from the test runs, and while your table is good, it is pretty wide. For example, the "STATE" field gets a lot of space, even though it only contains one character. Similar with PID and CPU, which are really wide. The benefit of `print_table()` is that it will determine appropriate column widths for you, so that the table will get automatically sized.

Here's a suggested rewrite which could use it:

```python
    rows = [["LAST_ARRIVAL", "ST", "PID", "TASK", "CPU", "COMMAND"]]
    tasks = list(tasks)
    tasks.sort(key=task_lastrun2now)
    for t in tasks:
        cpu = task_cpu(t)
        pid = get_pid(t)
        state = task_state_to_char(t)
        command = get_command(t)
        time_nanosec = task_lastrun2now(t)
        last_arrival = format_nanosecond_duration(time_nanosec)
        rows.append([last_arrival, state, pid, hex(t.value_()) cpu, command])
    print_table(rows)
```

drgn_tools/task.py
```diff
@@ -161,3 +170,91 @@ def task_cpu(task: drgn.Object) -> int:
...
+    return escape_ascii_string(task.comm.string_())
+
+
+def show_tasks_last_runtime(prog: Program, tasks: Iterable[Object]) -> None:
+    """
+    Display task information in their last arrival order.
+    """
+    print(
+        "LAST_ARRIVAL".ljust(25)
+        + "STATE".ljust(18)
```
brenns10, line 206: It would be nice to make this `ST` so that the column could be narrower

drgn_tools/task.py
```diff
@@ -161,3 +170,91 @@ def task_cpu(task: drgn.Object) -> int:
...
+        + "TASK".ljust(20)
+        + "CPU".ljust(15)
+        + "COMMAND".ljust(20)
+    )
+    tasklist = []
+    for t in tasks:
+        cpu = task_cpu(t)
+        pid = get_pid(t)
+        state = task_state_to_char(t)
+        task = find_task(prog, pid)
```
brenns10, line 217: This is unnecessary, `t` is already the task struct, you can simply use `hex(t.value_())` below.

brenns10 (changes requested):
One new comment, and a couple things I missed. It looks like a good start for the PS command, and I'm looking forward to merging it!

brenns10:
Also, please do re-push your branch to the internal gitlab to run vmcore tests as well. Thanks!

brenns10 (approved):
Looks good, here's a sample output pasted from the litevm test, for posterity:
```
   LAST_ARRIVAL     ST  PID  TASK                CPU  COMMAND          
  00 00:00:00.038  I   10   0xffff9bf681253180  0    kworker/0:1H     
  00 00:00:00.040  R   144  0xffff9bf682696300  0    pytest           
  00 00:00:00.281  I   17   0xffff9bf6812798c0  0    kworker/0:1      
  00 00:00:00.434  S   26   0xffff9bf6812d6300  0    kcompactd0       
  00 00:00:00.539  S   16   0xffff9bf68127ca40  0    migration/0      
  00 00:00:01.179  I   9    0xffff9bf681250000  0    kworker/u2:0     
  00 00:00:04.992  S   28   0xffff9bf6f76e98c0  0    khugepaged       
  00 00:00:23.022  I   15   0xffff9bf68125ca40  0    rcu_sched        
  00 00:00:28.314  S   140  0xffff9bf6826a6300  0    jbd2/vda-8       
  00 00:00:29.722  S   14   0xffff9bf68125e300  0    ksoftirqd/0      
  00 00:00:34.037  S   2    0xffff9bf681224a40  0    kthreadd         
  00 00:00:34.036  I   152  0xffff9bf6f75e98c0  0    dio/vda    
```

PR #32, opened 2023-11-28
numastat: Add numastat corelens module

jianfenw:
This commit implements numastat corelens module that enables users to dump detailed memory statistics for each NUMA node. The output is similar to 'numastat -m', which is to show meminfo-like system-wide memory usage.

This module supports UEK 4, 5, 6, and 7 for both x86-64 and aarch64. It is tested for all above settings. For each case, this module's output is compared against the output of `numastat -m`. Results match closely with only small differences.


jianfenw:
@brenns10 This PR has passed all live-vm and vmcore tests and is ready for review.
I would be grateful if you could take a look. Thanks!

drgn_tools/numastat.py
```diff
@@ -0,0 +1,280 @@
...
+            zone_stats["NR_MANAGED_PAGES"] = zone.managed_pages.value_()
+
+        all_zones_stats.append(zone_stats)
+
+    for zone_stats in all_zones_stats:
+        for name, value in zone_stats.items():
+            if name not in stats:
+                stats[name] = value
+            else:
+                stats[name] += value
```
brenns10, line 70: You can simplify this a bit by using `get` which lets you set a default.
```
stats[name] = stats.get(name, 0) + value
```

brenns10 (approved):
I had one nit on this, but honestly, it's not worth the added round trip, so I guess it's just an FYI for next time :)
Very nice work, looks ready to merge!

PR #33, opened 2023-11-29
Display loaded modules, their dependencies and parameter set

apalakun:


drgn_tools/lsmod.py
```diff
@@ -0,0 +1,165 @@
...
+            "PARAMETER".ljust(25),
+            "ADDRESS".rjust(15),
+            "TYPE".rjust(25),
+            "VALUE".rjust(10),
+        )
+        ker_parm = hex(use.kp.value_())
+        for i in range(0, use.num_kp.value_()):
+            parm_valf = ""
+            parm_val = ""
+            kp = Object(prog, "struct kernel_param", address=int(ker_parm, 16))
```
brenns10, line 84: It looks like you're having a little trouble iterating over this array. You can do it like this, and not need to maintain your manual ker_parm variable:
```suggestion
        for i in range(use.num_kp):
            parm_valf = ""
            parm_val = ""
            kp = use.kp[i]
```

drgn_tools/lsmod.py
```diff
@@ -0,0 +1,165 @@
...
+            try:
+                param_type = str(
+                    (
+                        (
+                            (identify_address(kp.ops.get).split(":")[1]).split(
+                                "+"
+                            )[0]
+                        ).split("param_")[1]
+                    )
+                )
```
brenns10, line 94: I think what you're trying to do is just get the symbol name corresponding to the `get()` function in the ops struct, which is a great strategy! The identify_address() helper is overkill for this. You can use drgn's built-in symbol lookup function (Program.symbol).

```python
try:
    param_type = prog.symbol(kp.ops.get).name
except LookupError:
    continue  # maybe log or print a warning?
if param_type.startswith("param_get_"):
    param_type = param_type[len("param_get_"):]
```

This leaves you with types like charp, string, uint, etc. And it also allows you to get rid of the big try/catch: you should only need it for this part.

drgn_tools/lsmod.py
```diff
@@ -0,0 +1,165 @@
...
+                if "charp" in param_type:
+                    param_type = "charp"
+                    parm_val = str(prog.read(kp.arg.value_(), 2).hex())
+                    if parm_val != "0000":
+                        parm_valf = str(int(parm_val, 16))
+                    elif parm_val == "0000":
+                        parm_valf = "NULL"
+                    else:
+                        parm_valf = "UNKNOWN Str"
```
brenns10, line 103: Reading the code, `param_get_charp()` interprets `kp->arg` as a `char **`, and reads that pointer. You would want something like this:
```python
value = Object(prog, "char **", value=kp.arg)
if value[0]:
    string = escape_ascii_string(value[0].string_())
else:
    string = "(null)"
```

drgn_tools/lsmod.py
```diff
@@ -0,0 +1,165 @@
...
+                elif "int" in param_type:
+                    parm_val = prog.read(kp.arg.value_(), 2).hex()
+                    if len(parm_val) > 2:
+                        parm_valf = str(get_parm_valf(prog, parm_val, kp))
+                    else:
+                        parm_valf = str(
+                            int(str(prog.read(kp.arg.value_(), 2).hex()), 16)
+                        )
+                    if str(parm_val) == "ffff":
+                        parm_valf = "-1"
```
brenns10, line 121: These integer types are defined something like this:
```
236  STANDARD_PARAM_DEF(byte,	unsigned char,		"%hhu", kstrtou8);
237  STANDARD_PARAM_DEF(short,	short,			"%hi",  kstrtos16);
238  STANDARD_PARAM_DEF(ushort,	unsigned short,		"%hu",  kstrtou16);
239  STANDARD_PARAM_DEF(int,		int,			"%i",   kstrtoint);
240  STANDARD_PARAM_DEF(uint,	unsigned int,		"%u",   kstrtouint);
241  STANDARD_PARAM_DEF(long,	long,			"%li",  kstrtol);
242  STANDARD_PARAM_DEF(ulong,	unsigned long,		"%lu",  kstrtoul);
243  STANDARD_PARAM_DEF(ullong,	unsigned long long,	"%llu", kstrtoull);
```

I think the way to handle this is to map these type names to their corresponding C types, and let drgn handle it. You can create the mapping like this (outside the loop):

```
# From STANDARD_PARAM_DEF in kernel/params.c
integer_params = {
    "byte": "unsigned char *",
    "short": "short *",
    "ushort": "unsigned short *",
    ...
}
```

Then, you can use that mapping in this if statement:
```
...
elif param_type in integer_params:
    value = cast(integer_params[param_type], kp.arg).value
```

drgn_tools/lsmod.py
```diff
@@ -0,0 +1,165 @@
...
+    :returns: int after converting from hex
+    """
+    parm_val1 = ""
+    for i in range(0, len(parm_val), 2):
+        parm_val1 = (
+            str(prog.read(kp.arg.value_(), 2).hex())[i : i + 2] + parm_val1
+        )
+    return int(str(parm_val1), 16)
+
+
```
brenns10, line 29: I'm not really sure I understand the purpose of this function, but I don't think we need it - see my approach for reading integers.

In general, even if you didn't use the approach of casting with drgn, you might find the following functions from drgn helpful:

Program.read_u8
Program.read_u16
Program.read_u32
Program.read_u64

drgn_tools/lsmod.py
```diff
@@ -0,0 +1,165 @@
...
+                    if len(parm_val) > 2:
+                        parm_valf = str(get_parm_valf(prog, parm_val, kp))
+                    else:
+                        parm_valf = str(
+                            int(str(prog.read(kp.arg.value_(), 2).hex()), 16)
+                        )
+                    if str(parm_val) == "ffff":
+                        parm_valf = "-1"
+                elif "bool" in param_type:
+                    parm_valf = "Y" if parm_val == 1 else "N"
```
brenns10, line 123: You could do mostly the same thing I suggested with integer types, but just use `cast("bool *", kp.arg)` and convert that to a bool the way you have it.

drgn_tools/lsmod.py
```diff
@@ -0,0 +1,165 @@
...
+                elif "ulong" in param_type:
+                    parm_valf = str(
+                        int(str(prog.read(kp.arg.value_(), 8).hex()), 16)
+                    )
+                elif "array" in param_type:
+                    parm_valf = str(
+                        int(str(prog.read(kp.arg.value_(), 1).hex()), 16)
+                    )
```
brenns10, line 131: Arrays need to have somewhat special handling. In those cases, `kp.arr` points to a `struct kparam_array`, which contains _another_ ops pointer, and info about the array size and element size.

drgn_tools/lsmod.py
```diff
@@ -0,0 +1,165 @@
...
+                ######
```
brenns10, line 132: You're missing support for `param_get_string()`, we'll need that too. It has a `struct kparam_string` structure. It's probably as simple as doing
```python
escape_ascii_string(kp.str.string.string_())
```

tests/test_lsmod.py
```diff
@@ -0,0 +1,11 @@
+# Copyright (c) 2023, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
+from drgn import ProgramFlags
+
+import drgn_tools.lsmod as lsmod
+
+
+def test_lsmod(prog):
+    if ProgramFlags.IS_LIVE & prog.flags:
+        return
```
brenns10, line 10: What concrete issues have you experienced that required you to skip tests on live kernels?

Our best and quickest tests run on live kernels, it's definitely preferred to run as many tests on the live kernels as possible, unless there's some issue that makes it impractical or infeasible. I can't think of a reason why this wouldn't work, but I may be wrong.

So if there's no reason for this, please delete these lines.

brenns10 (changes requested):
This will be a quite useful helper, thank you.

I do actually believe that this would be even better suited for contributing directly to drgn. While drgn already has a basic lsmod feature (see `contrib/lsmod.py`), helpers that can load module parameters would be nice to contribute upstream. We can talk about more after you've gone through the suggestions here.

brenns10:
Not sure if you're ready for review yet: there are still some failures on the gitlab vmcore tests. Let me know when you're ready for another look!

drgn_tools/lsmod.py
```diff
@@ -0,0 +1,167 @@
...
+            "ulong": "unsigned long *",
+        }
+        for i in range(0, mod.obj.num_kp):
+            parm_valf = ""
+            parm_val = ""
+            kp = mod.obj.kp[i]
+            try:
+                param_type = prog.symbol(kp.ops.get).name
+            except LookupError:
+                continue
```
brenns10, line 66: In the case of a LookupError, I think we can still print out the name, and indicate that we didn't know what to do with it. EG:

```python
table_value.append([kp.name.string_().decode("utf-8"), kern_parm, "UNKNOWN", "UNKNOWN"])
continue
```

This way, module parameters are not skipped over just because we had trouble with them.

drgn_tools/lsmod.py
```diff
@@ -0,0 +1,167 @@
...
+                param_type = param_type[len("param_get_") :]
+                if param_type in integer_params:
+                    parm_valf_address = cast(
+                        integer_params[param_type], kp.arg
+                    )
+                    try:
+                        parm_valf = parm_valf_address[0].value_()
+                    except FaultError:
+                        parm_valf = "(page fault)"
+                        continue
```
brenns10, line 78: Did you encounter page faults with integer types? I wouldn't expect that. If you're not working around a known issue, I'd prefer you not catch this error.

Regardless, with the `continue` statement here, the parameter will get skipped and won't be added to the list in the `FaultError` case. We should be including all parameters even on error. It would be best to remove the `continue` statement.

drgn_tools/lsmod.py
```diff
@@ -0,0 +1,167 @@
...
+                    try:
+                        if escape_ascii_string(kp.str.string.string_()):
+                            parm_valf = escape_ascii_string(
+                                kp.str.string.string_()
+                            )
+                        else:
+                            parm_valf = "(null)"
+                    except FaultError:
+                        parm_valf = "(page fault)"
+                        continue
```
brenns10, line 107: Remove the continue here as well.

drgn_tools/lsmod.py
```diff
@@ -0,0 +1,167 @@
...
+            elif "array" in param_type:
+                param_type = "array"
+                parm_valf = str(
+                    int(str(prog.read(kp.arg.value_(), 1).hex()), 16)
+                )
```
brenns10, line 112: A couple issues here.

- What you're doing is essentially reading a single byte pointed at by `kp.arg` and using that as the value of the array. That doesn't really make sense to me. For an array, you need to use `kp.arr` and look at the element type, and read each element like you would in the above cases. If you don't want to go to the trouble of reading arrays, that would be ok, but you should print a placeholder value. This one byte is meaningless and confusing in the table output.
- The pattern of `int(str(prog.read(...).hex()), 16)`... I understand what you're trying to do. `prog.read()` returns a `bytes` value and you're converting it to an int type. But using a string as an intermediary is not the correct way. Drgn has purpose built functions for reading different integer types. You could be using `prog.read_u8()` or `prog.read_u16()` ... [docs](https://drgn.readthedocs.io/en/latest/api_reference.html#drgn.Program.read_u8).

drgn_tools/lsmod.py
```diff
@@ -0,0 +1,167 @@
...
+            table_value.append(
+                [
+                    str(kp.name.string_().decode("utf-8")),
+                    str(ker_parm),
+                    param_type,
+                    str(parm_valf),
+                ]
+            )
+            ker_parm = hex(int(ker_parm, 16) + int(40))
```
brenns10, line 121: I'm assuming that 40 is the size of `struct kernel_param` and you're adjusting the hex value to point to the next parameter?

This is contrary to the entire point of drgn: it knows the size of structures, and it can tell you the address of objects. Rather than maintaining your own variable, and hard-coding structure sizes, let's have drgn maintain everything for us.

Please delete the `kpr_parm` variable entirely. When you need the hex address of the `struct kernel_param`, you can do `hex(kp.address_of_())`.

brenns10 (changes requested):

drgn_tools/lsmod.py
apalakun: Yes we have encountered and below is the localtest done on the vmcore 
[apalakun@lse-india-drgn drgn-tools]$ python3 -m drgn_tools.corelens /drgn_vmcore/githubCI/rds-uek5/vmcore --debuginfo /drgn_vmcore/githubCI/rds-uek5/ -M lsmod > ~/4.txt
error: Encountered exception in lsmod:
Traceback (most recent call last):
  File "/drgn_home/apalakun/github/drgn-tools/drgn_tools/corelens.py", line 340, in _run_module
    mod.run(prog, args)
  File "/drgn_home/apalakun/github/drgn-tools/drgn_tools/lsmod.py", line 157, in run
    print_module_parameters(prog)
  File "/drgn_home/apalakun/github/drgn-tools/drgn_tools/lsmod.py", line 72, in print_module_parameters
    parm_valf = parm_valf_address[0].value_()
_drgn.FaultError: could not read memory from kdump: Cannot get page I/O address: Page not present: pte[477] = 0x0: 0xffffffffc05dd000

drgn_tools/lsmod.py
apalakun: added string "array_length" to the output showing the length of the array

MODULE NAME:    parport_pc                                    
PARAM COUNT: 5                                                
ADDRESS    : 0xffffffffc0327390                               
PARAMETER  ADDRESS             TYPE   VALUE                   
init_mode  0xffffffffc03268a0  charp  (null)                  
dma        0xffffffffc03268c0  array  array_length: 16        
irq        0xffffffffc03268e0  array  array_length: 16        
io_hi      0xffffffffc0326900  array  array_length: 17        
io         0xffffffffc0326920  array  array_length: 17

brenns10:
Ok, I think the logic for the module parameter parsing all looks sound. However, I think the `print_module_parameters()` function needs to be broken down:

* It's got a ton of ifs, multiple loops, and several try/except statements, which make the nesting very deep, and the code more difficult to understand.
* It would be nice to add support for properly reading the array value out, but that would add even more code to an already large function.
* It seems like there should be a programmer's API that allows us to fetch the module parameter information, without printing it. As it is now, your only option is to print the parameters.

I pushed a commit to my own branch which modifies your implementation and addresses those issues, just to illustrate my point. I tried to simplify things as much as possible, and I split it out into three main functions:

1. `decode_param()` interprets a single kernel parameter and gets the drgn object for it, if possible
2. `module_params()` returns a dictionary of parameters for a single module (you can also use `KernelModule.params()`)
3. `print_module_parameters()` is much smaller now, all it is responsible for is formatting the data

[Here](https://github.com/brenns10/drgn-tools/commit/319b1fae0abac387a5e7bf13bb4a62340432d075) is my patch. We can just use that as-is, or if you want to incorporate the ideas yourself in your own way, that's great too. Just let me know.

brenns10:
From email discussion, we'll merge this and apply my patch on top.

brenns10 (approved):

PR #34, opened 2023-11-29
debuginfo: kernel modules with hyphen in filename

brenns10:
The module debuginfo code is already written to be careful with respect to module names that have hyphens and underscores. The filenames may not match the name that the kernel uses (e.g. filenames may contain hyphens, while the kernel's name contains underscores).

However, we were clearly not careful enough. When kernel module debuginfo is installed directly on the system, the loader did not search for the hyphenated version of module names with underscores. This led to missing debuginfo for modules like "dm_mod" (whose filename is dm-mod.ko.debug). Notably, the error did not occur on systems where drgn-tools extracts the RPM, since we take pains to rename all modules so their filename uses underscores instead of hyphens. The result is that this bug only really occurs when the debuginfo package is installed on the system.

Update the _find_debuginfo() function to allow hyphens or underscores, and check for the existence of both.

### DIFF
--- a/drgn_tools/debuginfo.py
+++ b/drgn_tools/debuginfo.py
@@ -310,19 +310,22 @@ def _get_debuginfo_paths(
 
 
 def _find_debuginfo(paths: List[Path], mod: str) -> Optional[Path]:
+    replace_pat = re.compile(r"[_-]")
     for search_dir in paths:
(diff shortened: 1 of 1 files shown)


biger410 (approved):
Looks good.

PR #44, opened 2023-12-06
task: add "ps" corelens module

SiddhiK29:


drgn_tools/task.py
```diff
@@ -215,6 +300,38 @@ def show_tasks_last_runtime(tasks: Iterable[Object]) -> None:
...
+                cpu,
+                hex(t.value_()),
+                state,
+                pct_mem,
+                vmem,
+                rss,
+                command,
+            ]
+        )
+    print_table(rows)
```
brenns10, line 332: This could be a matter of taste, but it seems like you could do this a little more concisely with something like:

```
rows.append(
    [
        get_pid(t),
        get_ppid(t),
...
```
Not everything needs to have a variable first.

drgn_tools/task.py
```diff
@@ -234,4 +351,4 @@ def run(self, prog: Program, args: argparse.Namespace) -> None:
         if args.last_run:
             show_tasks_last_runtime(for_each_task(prog))
         else:
-            raise NotImplementedError("currently, only ps -m is implemented")
+            show_taskinfo(prog, for_each_task(prog))
```
brenns10, line 354: So, this isn't quite everything in terms of the `ps` command we use in crash. And don't get me wrong, I don't think it needs to be identical, or contain every option. But there are some core things that would be nice. In particular:

* allow passing `pid | task | command` to filter by these. So you could do `ps java` to see all java process, or `ps 1` to see all threads of init, etc.
* `-k` filter to only kernel threads, or `-u` filter to only user threads
* `-G` display only the thread group leader (i.e task where `pid == ppid`)

The nice thing is that these can all be implemented in a straightforward way that allows everything to combine nicely:

```
tasks = for_each_task(prog)

if args.user:
    tasks = filter(tasks, is_user)
elif args.kernel:
    tasks = filter(tasks, is_kthread)
elif args.group_leader:
    tasks = filter(tasks, is_group_leader)

if args.last_run:
    show_tasks_last_runtime(tasks)
else:
    show_taskinfo(prog, tasks)
```

And this could also be extended to support some of the other ps options, if you wanted.

brenns10 (changes requested):
This looks great! Just missing a few features that I enumerated in the comments.
Thanks!

drgn_tools/task.py
```diff
@@ -197,6 +197,91 @@ def get_command(task: Object) -> str:
...
+                    task_anonrss = 0
+
+                task_shmrss = gtask.rss_stat.count[3].value_()
+                if task_shmrss < 0:
+                    task_shmrss = 0
+
+                rss_task += task_filerss + task_anonrss
+
+            except Exception:
+                rss_task = 0
```
brenns10, line 248: Multiple questions:

1. Iterating over every task in the system, just to find tasks of the same thread group, is inefficient. There is a field in `task_struct` called `thread_group` which appears to be a linked list of the threads. Can you use that instead?
2. Even then, iterating over every task in a thread group (i.e. process) could be quite expensive for processes with tons of threads. I think it might be worth adding a third argument to `get_rss()`, which would be a dict mapping `tgid` to `rss_task`. You would check this dictionary before computing it, and after computing a new value, you'd store it in the cache. This would ensure that you don't end up in an `O(N^2)` situation where you have a task with 1000 threads, and each time you do `get_rss()` on a thread you redo this, resulting in 1,000,000 iterations of the loop altogether.
3. Finally, what is the `except Exception` for? Which errors did you catch and why? Silently catching exceptions and continuing is not good for maintainable software. If there's a specific error that happens, it makes sense to catch that, but a blanked `except Exception` is a very wide net to cast, and will allow bugs to slip through.

Something like this maybe:
```suggestion
    gid = task.tgid.value_()
    if gid in rss_task_cache:
        rss_task = rss_task_cache[gid]
    else:
        rss_task = 0
        for gtask in list_for_each_entry("struct task_struct", 
                                                                task.thread_group.address_of_(),
                                                                "thread_group"):
            # note: this iteration doesn't include "task" itself, probably
            # need to add code to handle that.
            task_filerss = gtask.rss_stat.count[0].value_()
            if task_filerss < 0:
                task_filerss = 0

            task_anonrss = gtask.rss_stat.count[1].value_()
            if task_anonrss < 0:
                task_anonrss = 0

            task_shmrss = gtask.rss_stat.count[3].value_()
            if task_shmrss < 0:
                task_shmrss = 0

            rss_task += task_filerss + task_anonrss
        rss_task_cache[gid] = rss_task
```

drgn_tools/task.py
```diff
@@ -197,6 +197,91 @@ def get_command(task: Object) -> str:
...
+        anonrss = task.mm.rss_stat.count[1].counter.value_()
+        if anonrss < 0:
+            anonrss = 0
+        shmrss = task.mm.rss_stat.count[3].counter.value_()
+        if shmrss < 0:
+            shmrss = 0
+        rss_mm = filerss + anonrss
+
+    except Exception:
+        rss_mm = 0
```
brenns10, line 224: `shmrss` is unused... why collect it? same below as well.

drgn_tools/task.py
```diff
@@ -197,6 +197,91 @@ def get_command(task: Object) -> str:
...
+        filerss = task.mm.rss_stat.count[0].counter.value_()
+        if filerss < 0:
+            filerss = 0
+        anonrss = task.mm.rss_stat.count[1].counter.value_()
+        if anonrss < 0:
+            anonrss = 0
+        shmrss = task.mm.rss_stat.count[3].counter.value_()
+        if shmrss < 0:
+            shmrss = 0
+        rss_mm = filerss + anonrss
```
brenns10, line 221: The indexes you're using... 0, 1, 3. They are named enums in the kernel:
```c
43  enum {
44  	MM_FILEPAGES,	/* Resident file mapping pages */
45  	MM_ANONPAGES,	/* Resident anonymous pages */
46  	MM_SWAPENTS,	/* Anonymous swap entries */
47  	MM_SHMEMPAGES,	/* Resident shared memory pages */
48  	NR_MM_COUNTERS
49  };
```

It would be better to use `prog.constant("MM_FILEPAGES")` instead of "1", just for the principle of it. It helps the reader understand the magic number. Also, by doing `prog.constant("MM_SHMEMPAGES")`, you could detect whether that constant exists (it doesn't on UEK4) which may have been one of the exceptions you were trying to catch with the `except Exception`.

drgn_tools/task.py
```diff
@@ -197,6 +197,91 @@ def get_command(task: Object) -> str:
...
+        rss = get_rss(prog, task)
+        page_shift = prog["PAGE_SHIFT"].value_()
+        page_offset_base = prog["page_offset_base"].value_()
+        high_mem = prog["high_memory"].value_()
+        total_pages = ((high_mem - page_offset_base) >> page_shift) * 4
+        mem = (rss * 100) / total_pages
+        pct_mem = "%.1f" % mem
+    except Exception:
+        pct_mem = str(0.0)
+    return pct_mem
```
brenns10, line 270: Again, what's the `except Exception` for? It needs to have a good documented reason, or a more specific exception with a reason.

And `str(0.0)` can just be written as `"0.0"`, a bit easier that way :)

drgn_tools/task.py
```diff
@@ -197,6 +197,91 @@ def get_command(task: Object) -> str:
...
+def get_vmem(prog: Program, task: Object) -> float:
+    """
+    returns VSZ value
+    """
+    try:
+        page_size = prog["PAGE_SIZE"].value_()
+        vmem = (task.mm.total_vm.value_() * page_size) // 1024
+    except Exception:
+        vmem = 0
+    return vmem
```
brenns10, line 282: Same on the `except Exception`

brenns10 (changes requested):

drgn_tools/task.py
pssatapathy-oracle: This is done to inline with Oracle crash tool. As discussed, we will make it similar to
main line Crash , which is more appropriate

drgn_tools/task.py
pssatapathy-oracle: Agree with this and need a little modification ,

drgn_tools/task.py
brenns10: Got it, thanks!

drgn_tools/task.py
brenns10: > Let create set gid as index while traversing "for each task"
> and get the values directly from index when needed

I think this is what I'm proposing? `rss_task_cache` is a dictionary indexed by `gid`. The first time a task group is encountered, the task rss is computed and stored, and it will be reused for each subsequent thread of the task group.

The idea is that the caller will create that dictionary and the cache will be used for each call to avoid recomputing the task rss value.

We certainly don't want to do this in the `for_each_task()` function because most callers don't need memory usage information. But maybe I don't understand what you're suggesting. If I'm misunderstanding, I'm happy to take a look at a pseudocode or mock-up of what you're proposing!

drgn_tools/task.py
SiddhiK29: Thanks Stephen for pointing out this. I have incorporated these changes in new code.

drgn_tools/task.py
SiddhiK29: Don't really need any check for errors here. RSS calculation part of the code will take care of it so I have removed that.

drgn_tools/task.py
```diff
@@ -185,6 +195,173 @@ def get_command(task: Object) -> str:
...
 
 
+def get_ppid(task: Object) -> int:
+    """
+    :returns: Parent PID of the task
+    """
+    return task.parent.pid.value_()
+
+
+def build_rss_global(prog: Program) -> None:
```
brenns10, line 205: Ok, I did a thorough read through of the RSS accounting code, and I think we're completely overdoing this.

---

The whole "task_rss" and "mm_rss" thing just exists because of a performance optimization in the kernel. If these counters were only maintained as counters in the `mm_struct`, then there could be serious contention updating these values for processes with many threads. `SPLIT_RSS_COUNTING` allows each `task_struct` to contain a separate set of counters. For each task, every 64 page faults, the `mm_struct` gets its counters updated via `sync_mm_rss()`. This dramatically reduces the contention on the `mm_struct` counters.

In reality, we do not care about knowing a specific task's task_rss value. We _only_ care about the counters once you've combined the ones from the mm_struct and all of its associated task_structs. There's no need to cache any intermediate values. All we want is to have a single cache that maps an mm_struct to its final, computed RSS values.

There may be some value in keeping the filepages, anonpages, and shmempages separate. After all, `/proc/pid/status` breaks down these values too. I could see other helper code wanting to know this at some point. So let's try to keep the numbers separate, where possible.

And finally, there's one issue here: `build_rss_global()` computes data for every task in the system, which would be a waste. For an example, in the future, we would like to support an invocation like `ps <PID>` or `ps COMM`, which would only return rows for a subset of tasks. Computing the RSS for every single task would be a waste for those invocations.

Taking these points into account, here's a simplified and cleaned up implementation, maybe we could start with something like this instead?

```python
class TaskRss(NamedTuple):
    """Represent's a task's resident set size. See task_rss()."""

    file: int
    anon: int
    shmem: int

    @property
    def total(self) -> int:
        return self.file + self.anon + self.shmem


def task_rss(task: Object, cache: Optional[Dict[int, TaskRss]]) -> TaskRss:
    """
    Return the task's resident set size (RSS) in pages

    The task's RSS is the number of pages which are currently resident in
    memory. The RSS values can be broken down into anonymous pages (not bound to
    any file), file pages (those associated with memory mapped files), and
    shared memory pages (those which aren't associated with on-disk files, but
    belonging to shared memory mappings). This function returns a tuple
    containing each category, but the common behavior is to use the "total"
    value which sums them up.

    :param task: ``struct task_struct *`` for which to compute RSS
    :param cache: if provided, we can use this to cache the mapping of
      "mm_struct" to RSS. This helps avoid re-computing the RSS value for
      processes with many threads, but note that it could result in out of date
      values on a live system.
    :returns: the file, anon, and shmem page values
    """
    mmptr = task.mm.value_()
    if cache and mmptr in cache:
        return cache[mmptr]

    # Kthreads have a NULL mm, simply skip them, returning 0.
    if not task.mm:
        return TaskRss(0, 0, 0)

    prog = task.prog_
    MM_FILEPAGES = prog.constant("MM_FILEPAGES").value_()
    MM_ANONPAGES = prog.constant("MM_ANONPAGES").value_()
    try:
        MM_SHMEMPAGES = prog.constant("MM_SHMEMPAGES").value_()
    except LookupError:
        MM_SHMEMPAGES = -1

    # Start with the counters from the mm_struct
    file = anon = shmem = 0
    file += task.mm.rss_stat.count[MM_FILEPAGES].counter.value_()
    anon += task.mm.rss_stat.count[MM_ANONPAGES].counter.value_()
    if MM_SHMEMPAGES >= 0:
        shmem += task.mm.rss_stat.count[MM_SHMEMPAGES].counter.value_()

    # Augment the values with the ones from the task's rss_stat.
    # The task rss_stat is what gets updated directly, every so often these
    # changes get sync'd to the mm_struct.
    file += task.rss_stat.count[MM_FILEPAGES].value_()
    anon += task.rss_stat.count[MM_ANONPAGES].value_()
    if MM_SHMEMPAGES >= 0:
        shmem += task.rss_stat.count[MM_SHMEMPAGES].value_()

    # Finally, augment the values with the ones from the rest of the thread
    # group.
    for gtask in list_for_each_entry(
        "struct task_struct",
        task.thread_group.address_of_(),
        "thread_group",
    ):
        file += gtask.rss_stat.count[MM_FILEPAGES].value_()
        anon += gtask.rss_stat.count[MM_ANONPAGES].value_()
        if MM_SHMEMPAGES >= 0:
            shmem += gtask.rss_stat.count[MM_SHMEMPAGES].value_()

    rss = TaskRss(file, anon, shmem)
    if cache is not None:
        cache[mmptr] = rss
    return rss
```

The benefit of this `task_rss()` function is that it can be just as easily used as a helper function for other code. It works great on live systems (so long as you don't keep using the same `cache`), but it is efficient for the case of the `ps` command too.

drgn_tools/task.py
```diff
@@ -185,6 +195,173 @@ def get_command(task: Object) -> str:
...
+            # Prior to 021182e52fe0 ("x86/mm: Enable KASLR for physical mapping
+            # memory regions"), page_offset_base did not exist and there was
+            # just a constant for the beginning of the memory map.
+            page_offset_base = 0xFFFF880000000000
+
+        total_pages = (high_mem - page_offset_base) >> page_shift
+    else:
+        raise NotImplementedError(
+            "%MEM calculation is not (yet) implemented for aarch64"
+        )
```
brenns10, line 346: What if we used the function `totalram_pages()` instead of this architecture specific computation of `total_pages`?

This may not be exactly the same, but it strikes me that `high_mem` is a pretty archaic remnant of 32-bit kernels at this point.

drgn_tools/task.py
```diff
@@ -185,6 +195,173 @@ def get_command(task: Object) -> str:
...
+    mem = (tot_rss * 100) / total_pages
+    pct_mem = "%.1f" % mem
+
+    return pct_mem
```
brenns10, line 350: Please return the value as a `float`, and let the caller decide how to format it. It doesn't make sense for `get_pct_mem()` to return a string.

drgn_tools/task.py
```diff
@@ -185,6 +195,173 @@ def get_command(task: Object) -> str:
...
+    rss = rss_mm + rss_task
+    page_size = prog["PAGE_SIZE"].value_()
+    rss_kb = (rss * page_size) // ByteToKB
+
+    return rss_kb
+
+
+def get_pct_mem(prog: Program, task: Object) -> str:
+    """
+    :returns %MEM value
```
brenns10, line 323: If you're going to have a one-line docstring, it should be like the following:

```suggestion
    Return the percent memory usage value
```

The `:returns:` keyword can be used in multiline docstrings to explicitly document a return value (but then it would need both colons).

drgn_tools/task.py
```diff
@@ -185,6 +195,173 @@ def get_command(task: Object) -> str:
...
+        rss_task_dict[gid] = rss_task
+
+    rss = rss_mm + rss_task
+    page_size = prog["PAGE_SIZE"].value_()
+    rss_kb = (rss * page_size) // ByteToKB
+
+    return rss_kb
+
+
+def get_pct_mem(prog: Program, task: Object) -> str:
```
brenns10, line 321: This function doesn't need to accept `prog` as an argument. Please remove it. You can put the following line at the beginning of the function:
```python
prog = task.prog_
```

drgn_tools/task.py
```diff
@@ -185,6 +195,173 @@ def get_command(task: Object) -> str:
...
+        raise NotImplementedError(
+            "%MEM calculation is not (yet) implemented for aarch64"
+        )
+    mem = (tot_rss * 100) / total_pages
+    pct_mem = "%.1f" % mem
+
+    return pct_mem
+
+
+def get_vmem(prog: Program, task: Object) -> float:
```
brenns10, line 353: Again, don't pass `prog` here, get it from `task`.

brenns10 (changes requested):

drgn_tools/task.py
SiddhiK29: I had tested with `totalram_pages()` initially but as you said the value will not be exactly same, it will be bit different like below:

total_pages = 4456448 ( with above implementation)
```
4124985  4114326  0    0xffff91b3e199b000  S   1.6   1711276  290144  osms-agent
4127087  4114326  0    0xffff91b5b3161800  S   1.6   1711276  290144  osms-agent
4129199  4114326  2    0xffff91b430f6c800  S   1.6   1711276  290144  osms-agent
4131335  4114326  0    0xffff91b66f92e000  S   1.6   1711276  290144  osms-agent
```

totalram_pages = 4022584
```
4124985  4114326  0    0xffff91b3e199b000  S   1.8   1711276  290144  osms-agent
4127087  4114326  0    0xffff91b5b3161800  S   1.8   1711276  290144  osms-agent
4129199  4114326  2    0xffff91b430f6c800  S   1.8   1711276  290144  osms-agent
4131335  4114326  0    0xffff91b66f92e000  S   1.8   1711276  290144  osms-agent
```

If this much difference in the output is fine then we can use totalram_pages() function. Please let me know your thought on this.

brenns10: I took a second look and still think `totalram_pages()` is the best way. I skimmed the code from a couple locations to ensure I'm correct.

1. The kernel uses the same logic as `totalram_pages()` to report `MemTotal` in `/proc/meminfo`.
2. The userspace `ps(1)` utility (as implemented by procps-ng) divides the process RSS by the `MemTotal` value to get `%MEM`.
3. Crash has a more convoluted path toward getting the total memory, I didn't completely follow it during my brief skim.

As far as I'm concerned, `ps(1)` is the ground truth here, not crash. So let's go with `totalram_pages()` and accept the difference from crash!

SiddhiK29: okay, I have made changes in latest code. Thanks!

drgn_tools/task.py
SiddhiK29: Thanks for the detailed explanation! We have implemented this in latest code.

drgn_tools/task.py
```diff
@@ -185,6 +191,133 @@ def get_command(task: Object) -> str:
...
+    :param cache: if provided, we can use this to cache the mapping of
+      "mm_struct" to RSS. This helps avoid re-computing the RSS value for
+      processes with many threads, but note that it could result in out of date
+      values on a live system.
+    :returns: the file, anon, and shmem page values
+    """
+    try:
+        mmptr = task.mm.value_()
+        if mmptr is None:
+            return TaskRss(0, 0, 0)
```
brenns10, line 237: `mmptr` can never be `None`, so this is dead code and can be removed. If `mmptr` is NULL, it will become the integer value zero when you call `.value_()`.

drgn_tools/task.py
```diff
@@ -185,6 +191,133 @@ def get_command(task: Object) -> str:
...
+    except AttributeError:
+        return TaskRss(0, 0, 0)
```
brenns10, line 239: In what case does an `AttributeError` occur? I believe all the way back to UEK4 there is a `mm` field within `task_struct`. It's a quite important field.

I don't think there needs to be a "try/except" block here at all.

drgn_tools/task.py
```diff
@@ -185,6 +191,133 @@ def get_command(task: Object) -> str:
...
+    except LookupError:
+        MM_SHMEMPAGES = -1
+
+    filerss = anonrss = shmemrss = 0
+
+    if mmptr:
+        filerss = task.mm.rss_stat.count[MM_FILEPAGES].counter.value_()
+        anonrss = task.mm.rss_stat.count[MM_ANONPAGES].counter.value_()
+        if MM_SHMEMPAGES >= 0:
+            shmemrss = task.mm.rss_stat.count[MM_SHMEMPAGES].counter.value_()
```
brenns10, line 259: mmptr is null when a task is a kthread. In that case, the memory usage is 0 by definition: the kthread has no separate address space, and there's not even any point in looking at the rss_stat in the task_struct, nor iterating over members of the group.

We should get rid of this if-check here, and simply include near the very top of the function:
```python
if not mmptr:
    return TaskRss(0, 0, 0)
```

drgn_tools/task.py
```diff
@@ -185,6 +191,133 @@ def get_command(task: Object) -> str:
...
+def get_ppid(task: Object) -> int:
+    """
+    :returns: Parent PID of the task
+    """
+    return task.parent.pid.value_()
+
+
+class TaskRss(NamedTuple):
+    """
+    : Represent's a task's resident set size. See task_rss().
```
brenns10, line 203: No need to have the colon at the beginning of this line.

Also it would be nice to mention that all the values are in units of pages. I clearly forgot to write that down :)

drgn_tools/task.py
```diff
@@ -185,6 +191,133 @@ def get_command(task: Object) -> str:
...
+    rss_file: int
+    rss_anon: int
+    rss_shmem: int
+
+    @property
+    def total(self) -> int:
+        return self.rss_file + self.rss_anon + self.rss_shmem
+
+
+def get_task_rss(task: Object, cache: Optional[dict] = None) -> TaskRss:
```
brenns10, line 215: The type annotation of cache should be: `Optional[Dict[int, TaskRss]]`. 
You'll need `from typing import Dict`.

drgn_tools/task.py
```diff
@@ -185,6 +191,133 @@ def get_command(task: Object) -> str:
...
+
+    return rss
+
+
+def get_rss(task: Object) -> int:
+    """
+    Return the task's resident set size (RSS) in kilobytes
+    """
+    prog = task.prog_
+    task_rss = get_task_rss(task).total
```
brenns10, line 289: Calling `get_task_rss(task)` without providing a cache parameter defeats the entire purpose of having it. Same goes for the call in `get_pct_mem()`. Only by providing the cache parameter can you ensure that the computation is efficient.

Honestly, I'd recommend deleting both helper functions and computing them directly in `show_taskinfo()`.

drgn_tools/task.py
```diff
@@ -203,6 +336,30 @@ def show_tasks_last_runtime(tasks: Iterable[Object]) -> None:
...
+                str(get_ppid(t)),
+                str(task_cpu(t)),
+                hex(t.value_()),
+                task_state_to_char(t),
+                str("%.1f" % get_pct_mem(t)),
+                str(get_vmem(t)),
+                str(get_rss(t)),
+                get_command(t),
+            ]
+        )
```
brenns10, line 359: In order to gain the benefit of caching with `get_task_rss()`, you need to create the cache here and pass it in to each function call. I'd suggest also getting rid of `get_rss()` and `get_pct_mem()` since they both need this data, and it's simpler to just fetch it once and compute it all here.
```suggestion
def show_taskinfo(prog: Program, tasks: Iterable[Object]) -> None:
    """
    Display task information.
    """
    rows = [["PID", "PPID", "CPU", "TASK", "ST", "%MEM", "VSZ", "RSS", "COMM"]]
    tasks = list(tasks)
    tasks.sort(key=get_pid)
    page_size = int(prog["PAGE_SIZE"])
    total_mem = int(totalram_pages(prog))
    rss_cache = {}
    for t in tasks:
        task_rss = get_task_rss(t, rss_cache)
        rss_kb = task_rss.total * page_size // ByteToKB
        pct_mem = task_rss.total * 100 / pct_mem
        rows.append(
            [
                str(get_pid(t)),
                str(get_ppid(t)),
                str(task_cpu(t)),
                hex(t.value_()),
                task_state_to_char(t),
                str("%.1f" % pct_mem),
                str(get_vmem(t)),
                str(rss_kb),
                get_command(t),
            ]
        )
```

brenns10 (changes requested):

brenns10 (approved):

brenns10:
Thanks for all the work on this! Looks good.

PR #45, opened 2023-12-08
Block dump dispatched io for sq

biger410:
- dump dispatched io from sq
- use crash style hung time for io pending

### DIFF
--- a/drgn_tools/block.py
+++ b/drgn_tools/block.py
@@ -20,6 +20,7 @@ from drgn.helpers.linux.list import list_for_each_entry
 from drgn_tools.bitops import for_each_bit_set
 from drgn_tools.corelens import CorelensModule
 from drgn_tools.util import has_member
+from drgn_tools.util import timestamp_str
 from drgn_tools.util import type_exists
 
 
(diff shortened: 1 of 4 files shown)


drgn_tools/block.py
```diff
@@ -216,21 +217,21 @@ def for_each_mq_pending_request(q: Object) -> Iterable[Tuple[Object, Object]]:
...
+    Get io pending time in ns
 
     :param rq: ``struct request *`` or ``struct request``
     :returns: request pending time
     """
     prog = rq.prog_
     if has_member(rq, "start_time"):
         # TODO: arch specific
-        return prog["jiffies_64"] - rq.start_time
+        return (prog["jiffies_64"] - rq.start_time).value_() * 1000000
```
brenns10, line 230: Since drgn v0.0.23, "jiffies" is a special-cased variable that Drgn maintains for us. Since you touched this code, please go ahead and remove the todo and change it to `prog["jiffies"]`.

brenns10 (changes requested):
One small request, otherwise it looks great, thanks!

drgn_tools/block.py
biger410: done

brenns10 (approved):

PR #37, opened 2023-11-30
Add block-info module

srivathsa729:
block-info prints scsihosts, scsi-device information and pending in-flight IOs

Authored-by: Gulam Mohamed <gulam.mohamed@oracle.com>
Co-authored-by: Srivathsa Dara <srivathsa.d.dara@oracle.com>

PR #46, opened 2023-12-11
Slabinfo

srivathsa729:
Add slabinfo module

Prints slab cache information


brenns10:
It looks like this contains your blockinfo module too, from #37. Could you keep the slabinfo in a separate branch, so we can review them independently?

srivathsa729:
sorry, my bad

PR #47, opened 2023-12-12
Add slabinfo module

srivathsa729:
Prints slab cache information

Authored-by: Partha Satapathy <partha.satapathy@oracle.com>

### DIFF
--- a/drgn_tools/slabinfo.py
+++ b/drgn_tools/slabinfo.py
@@ -1,0 +1,280 @@
+# Copyright (c) 2023, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
+"""
+Helper to view slabinfo data
+"""
+import argparse
+from typing import NamedTuple
+from typing import Set
+
+from drgn import cast
+from drgn import NULL
+from drgn import Object
+from drgn import Program
+from drgn.helpers.linux.cpumask import for_each_present_cpu
+from drgn.helpers.linux.list import list_for_each_entry
+from drgn.helpers.linux.nodemask import for_each_online_node
+from drgn.helpers.linux.percpu import per_cpu_ptr
+from drgn.helpers.linux.slab import _get_slab_cache_helper
+from drgn.helpers.linux.slab import for_each_slab_cache
+
+from drgn_tools.corelens import CorelensModule
+from drgn_tools.table import print_table
+
+
+class SlabCacheInfo(NamedTuple):
+    """Describes summary information about a slab cache"""
+
+    cache: Object
+    """The ``struct kmem_cache *`` object"""
+    objsize: int
+    """Size of each individual object"""
+    allocated: int
+    """Number of allocated objects"""
+    total: int
+    """Total count of objects"""
+    nr_slabs: int
+    """Number of slabs"""
+    ssize: int
+    """Slab size"""
+    name: str
+    """Name of the slab cache"""
+
+
+def kmem_cache_pernode(cache: Object, nodeid: int) -> tuple:
+    """
+    Get number of slabs, objects and partial object per node
+    traverse though the partial list of node
+    and count the number of partial objects
+
+    :param cache: `struct kmem_cache` drgn object
+    :param nodeid: node index
+    "returns: a tuple with node information
+    """
+    nr_slabs = 0
+    nr_total_objs = 0
+    nr_free = 0
+    nr_partial = 0
+    node_total_use = 0
+
+    kmem_cache_node = cache.node[nodeid]
+    x = 0
+
+    prog = cache.prog_
+
+    nr_slabs = kmem_cache_node.nr_slabs.counter.value_()
+    nr_total_objs = kmem_cache_node.total_objects.counter.value_()
+    nr_partial = kmem_cache_node.nr_partial.value_()
+
+    page = Object(prog, "struct page", address=0x0)
+    partial_slab_list = "slab_list" if hasattr(page, "slab_list") else "lru"
+
+    for page in list_for_each_entry(
+        "struct page", kmem_cache_node.partial.address_of_(), partial_slab_list
+    ):
+        nrobj = page.objects.value_()
+        nrinuse = page.inuse.value_()
+        x = nrobj - nrinuse
+        node_total_use += nrinuse
+        nr_free += x
+
+    return nr_slabs, nr_total_objs, nr_partial, node_total_use, nr_free
+
+
+def kmem_cache_percpu(cache: Object) -> int:
+    """
+    Count the number of cpu_slab pages for all nodes.
+
+    :param: `struct kmem_cache` drgn object
+    """
+
+    cpu_per_node = 0
+    prog = cache.prog_
+    cpu_slab_ptr = NULL
+
+    for cpuid in for_each_present_cpu(prog):
+        per_cpu_slab = per_cpu_ptr(cache.cpu_slab, cpuid)
+        cpu_slab_ptr = per_cpu_slab.page
+
+        if not cpu_slab_ptr:
+            continue
+        cpu_per_node += 1
+
+    return cpu_per_node
+
+
+def collect_node_info(cache: Object) -> tuple:
+    """
+    Parse through each node to collect per-node slab data
+
+    :param: `struct kmem_cache` drgn object
+    :returns: a tuple containing per-node slab data
+    """
+    nr_slabs = 0
+    nr_total_objs = 0
+    nr_free = 0
+    nr_partial = 0
+    node_total_use = 0
+
+    prog = cache.prog_
+
+    for node in for_each_online_node(prog):
+        slabs, total_objs, partial, total_use, free = kmem_cache_pernode(
+            cache, node
+        )
+        nr_slabs += slabs
+        nr_total_objs += total_objs
+        nr_partial += partial
+        node_total_use += total_use
+        nr_free += free
+
+    return nr_slabs, nr_total_objs, nr_partial, node_total_use, nr_free
+
+
+def slub_get_cpu_freelist_cnt(
+    cpu_freelist: Object, slub_helper: Object
+) -> int:
+    """
+    Get number of elements in percpu freelist
+
+    :param slab_cache: `struct kmem_cache` drgn object
+    :param cpu_freelist: `void**` pointer to next available object
+    :param slub_helper: slab cache helper object
+    :returns: the count of per cpu free objects
+    """
+    cpu_free_set: Set[int] = set()
+    slub_helper._slub_get_freelist(cpu_freelist, cpu_free_set)
+
+    return len(cpu_free_set)
+
+
+def slub_per_cpu_partial_free(cpu_partial: Object) -> int:
+    """
+    Get the partial free from percpu partial list
+
+    :param cpu_partial: `struct page` drgn object
+        of kmem_cache->cpu_slab->partial list
+    :returns: free objects from partial list
+    """
+
+    partial_free = partial_objects = partial_inuse = 0
+
+    while cpu_partial:
+        page = cast("struct page *", cpu_partial)
+
+        partial_objects = page.objects.value_()
+        partial_inuse = page.inuse.value_()
+        partial_free += partial_objects - partial_inuse
+        cpu_partial = page.next
+
+    return partial_free
+
+
+def kmem_cache_slub_info(cache: Object) -> tuple:
+    """
+    For given kmem_cache object, parse through each cpu
+    and get number of total slabs and  free objects
+
+    :param: `struct kmem_cache` drgn object
+    :returns: total slabs, free objects
+    """
+    prog = cache.prog_
+
+    total_slabs = objects = free_objects = 0
+    slub_helper = _get_slab_cache_helper(cache)
+
+    for cpuid in for_each_present_cpu(prog):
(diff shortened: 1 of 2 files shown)


drgn_tools/slabinfo.py
```diff
@@ -0,0 +1,310 @@
...
+        partial_frees = slub_partial_free(prog, slab, cpu_partial)
+        free_objects += partial_frees
+        total_slabs += 1
+
+    collect_node_info(
+        prog, slab, nrnodeid, total_slabs, free_objects, total_inuse
+    )
+
+
+def get_slab_info(prog: drgn.Program) -> None:
```
brenns10, line 287: It might be nicer to rename this to `print_slab_info()`? Since this just prints the data, it doesn't return it.

drgn_tools/slabinfo.py
```diff
@@ -0,0 +1,310 @@
...
+            break
+        freelist_set.add(ptr)
+        ptr = try_hardened_freelist_dereference(
+            prog, slab_cache, (ptr + slab_cache.offset.value_())
+        )
+
+
+def slub_get_cpu_freelist(
+    prog: drgn.Program, slab_cache: drgn.Object, cpu_freelist: drgn.Object
+) -> set:
```
brenns10, line 181: This annotation should be `Set[int]`.

drgn_tools/slabinfo.py
```diff
@@ -0,0 +1,310 @@
...
+
+def get_slab_info(prog: drgn.Program) -> None:
+    """
+    Print the slab headers and initiate the slab collection
+    process for all slabs
+    """
+    output.append(
+        ["CACHE", "OBJSIZE", "ALLOCATED", "TOTAL", "SLABS", "SSIZE", "NAME"]
+    )
+    nrnodeid = prog["nr_node_ids"].value_()
```
brenns10, line 295: It might make sense to skip this, not provide the `nrnodeid` parameter, and just let the `collect_node_info()` function use [for_each_online_node](https://drgn.readthedocs.io/en/latest/helpers.html#drgn.helpers.linux.nodemask.for_each_online_node) to iterate over each relevant node.

drgn_tools/slabinfo.py
```diff
@@ -0,0 +1,310 @@
...
+    Collect the per-slab data.
+    Traverse though all the CPUs to collect the slab data.
+    """
+    total_slabs = objects = free_objects = total_inuse = 0
+
+    for cpuid in for_each_online_cpu(prog):
+        try:
+            per_cpu_slab = per_cpu_ptr(slab.cpu_slab, cpuid)
+        except (AttributeError, drgn.ObjectAbsentError):
+            continue
```
brenns10, line 251: I'm commenting here, but I'm curious about it at pretty much every use: in what cases do you encounter `ObjectAbsentError`? As far as I understand, this only occurs when reading values out of stack traces.

As for `AttributeError`, I can see that would happen if the type `kmem_cache` didn't contain a `cpu_slab` field - which is is the case for systems where slab is enabled rather than slub. While we don't need to _support_ slab, we should at least raise an understandable error, rather than silently ignoring it and crashing later.

I guess what I'm driving at is, why are we catching these exceptions? Without a documented explanation, I don't really like having exception handlers. They tend to end up suppressing errors we would want to see, and if that happens, the resulting behavior is usually incomplete in some way that would harm our bug analysis. Having the program crash is really unfortunate, and I don't like that either, but it would be better to do that than have flawed debugging analyses.

drgn_tools/slabinfo.py
```diff
@@ -0,0 +1,310 @@
...
+            objects = 0
+
+        free_objects += objects - page_inuse
+        cpu_free_objects = slub_get_cpu_freelist_cnt(prog, slab, cpu_freelist)
+        free_objects += cpu_free_objects
+        objects = 0
+        page_inuse = 0
+        partial_frees = slub_partial_free(prog, slab, cpu_partial)
+        free_objects += partial_frees
+        total_slabs += 1
```
brenns10, line 280: Does this account for the `cpu_partial` slabs?

drgn_tools/slabinfo.py
```diff
@@ -0,0 +1,310 @@
...
+            objects = 0
+            continue
+        if objects < 0:
+            objects = 0
+
+        free_objects += objects - page_inuse
+        cpu_free_objects = slub_get_cpu_freelist_cnt(prog, slab, cpu_freelist)
+        free_objects += cpu_free_objects
+        objects = 0
+        page_inuse = 0
```
brenns10, line 277: nit: no need to reset these here, they're set to zero, or the proper value at every iteration of the loop.

drgn_tools/slabinfo.py
```diff
@@ -0,0 +1,310 @@
...
+                nr_free += x
+        except LookupError:
+            for page in list_for_each_entry(
+                "struct page", kmem_cache_node.partial.address_of_(), "lru"
+            ):
+                nrobj = page.objects.value_()
+                nrinuse = page.inuse.value_()
+                x = nrobj - nrinuse
+                node_total_use += nrinuse
+                nr_free += x
```
brenns10, line 74: You're repeating the body of the loop for both conditions here. At some point, somebody is going to update this code and forget to update the other copy of it.

It would be better to have something like `for_each_partial_slab()` or something, which detects the proper way to iterate over the slabs, and returns the appropriate generator. That way you can avoid duplicating code.

drgn_tools/slabinfo.py
```diff
@@ -0,0 +1,310 @@
...
+
+
+def slub_partial_free(
+    prog: drgn.Program, slab_cache: drgn.Object, slab: drgn.Object
+) -> int:
+    """
+    Collect partial list from kmem_cache_cpu
+    """
+    cpu_partial = slab
+    pss_cnt = 0
```
brenns10, line 209: `pss_cnt` only seems to be incremented, never used. What is its purpose?

drgn_tools/slabinfo.py
```diff
@@ -0,0 +1,310 @@
...
+        except (AttributeError, drgn.ObjectAbsentError):
+            page_inuse = 0
+
+        try:
+            objects = cpu_slab_ptr.objects.value_()
+        except (AttributeError, drgn.ObjectAbsentError):
+            objects = 0
+            continue
+        if objects < 0:
+            objects = 0
```
brenns10, line 271: When does this happen, and why?

drgn_tools/slabinfo.py
```diff
@@ -0,0 +1,310 @@
...
+
+    if slab_freelist_hardened and needswab:
+        result = freelist_dereference_swab(prog, slab_cache, ptr_addr)
+    if not slab_freelist_hardened and not needswab:
+        try:
+            result = prog.read_word(ptr_addr)
+        except (AttributeError, drgn.ObjectAbsentError):
+            result = 0
+
+    return result
```
brenns10, line 157: 1. As far as I can tell, this ignores the `freelist_dereference_noswab()` case.
2. It looks like these parts of the code are adapted from the drgn slab code. In the drgn slab helpers there are comments regarding the different commits that cause us to need each of the variants. The drgn slab helpers also support `struct slab` rather than `struct page`, which we will need for UEK8.
3. This doesn't end up caching the `freelist_dereference()` choice, which means we end up redoing the checks for every time we dereference a freelist pointer.

I think we could solve all three, by simply using the `drgn.helpers.linux.slab._SlabCacheHelperSlub` class. You'll need to read the code to find it, but it provides most of this stuff in a way that's more compatible, more efficient, and better tested.

Essentially, inside of the `kmem_cache_info_slub()` function, you would do something like this:

```python
helper = _SlabCacheHelperSlub(cache)
```

On creation, `helper` will have a function `helper._slub_get_freelist()` which you can pass a freelist pointer into, and a set, and it will add each element of the freelist in, using the correct method.

I know that the downside here is that the `_SlabCacheHelperSlub` class is not a public API of drgn. But for one thing, each drgn-tools release is specific to a drgn release, so we know we won't get suddenly broken. For another, I think it would be worthwhile to try to make some of that API public, that's something we can discuss with Omar a bit.

brenns10 (changes requested):
Thank you for this, it's looking like a great start. The slab cache stuff is quite complex. I've got a good understanding of some things, but a lot of my review here is more focused on the design of the code. I think I'd like to ask @imran-kn to review the logic as well, before we go ahead.

drgn_tools/slabinfo.py
srivathsa729: Added that part of code from crash, but we have never encountered it. Removing this.

drgn_tools/slabinfo.py
srivathsa729: It's a debug variable, removing it

drgn_tools/slabinfo.py
```diff
@@ -0,0 +1,310 @@
...
+
+
+def slub_get_freelist(
```
imran-kn, line 160: We can use drgn's _slub_get_freelist for _SlabCacheHelperSlub

drgn_tools/slabinfo.py
```diff
@@ -0,0 +1,310 @@
...
+    """
+    cpu_free_set_rc = slub_get_cpu_freelist(prog, slab_cache, cpu_freelist)
+    return len(cpu_free_set_rc)
+
+
+def slub_partial_free(
+    prog: drgn.Program, slab_cache: drgn.Object, slab: drgn.Object
+) -> int:
+    """
+    Collect partial list from kmem_cache_cpu
```
imran-kn, line 206: This gives the notion that helper is returning a list. May be change it to something like:

"Get number of objects on per-cpu partial slab lists"

drgn_tools/slabinfo.py
```diff
@@ -0,0 +1,310 @@
...
+    """
+    Collect freelist data from kmem_cache_cpu and
+    return the count of free objects
+    """
+    cpu_free_set_rc = slub_get_cpu_freelist(prog, slab_cache, cpu_freelist)
+    return len(cpu_free_set_rc)
+
+
+def slub_partial_free(
+    prog: drgn.Program, slab_cache: drgn.Object, slab: drgn.Object
```
imran-kn, line 203: you can get "prog" from slab_cache.prog_ or slab.prog_. Also partial slabs can be both in per-cpu and per-node lists. Although the description indicates that it is using per-cpu slab list (i.e from kmem_cache_cpu) the name of the function itself should reflect that we are using per-cpu and not per node partial slab list here. I don't see slab_cache being used in this function so you can remove it. Or since this is supposed to find number of objects on per-cpu partial slabs for a give cache and cpu, you can avoid passing slab and just pass the slab_cache and cpu number

drgn_tools/slabinfo.py
imran-kn: +1

imran-kn (commented):
Besides the comments provided in the review, it looks like that the current organization of helpers is inspired by crash utility. 
With drgn the helpers can be structured in a better way. Something along these lines 

```
for each slab_cache:
    get info about per-cpu slabs:
        get info about lockless/per-cpu free list
        get info about active slab
        get info about partial (other than the active) slabs
   get info about other slabs:
         get info about per-node partial slabs
          get info about full slabs

We can dump the information thus obtained in a format similar to crash's format but having helpers organized like above will make the code more manageable and will allow us to use these helpers in other places as well.
Also make use of pre-existing slab helpers as much as possible, even if these are non-public

**Comment by srivathsa729** [comment]
The updated module is using drgn api as suggested, but one thing we found is that this approach is slow compared to previous version (which was using partha's own api).

When ran on following gitlab test vmcores:
smp_ipi-uek5, smp_ipi-uek6 and smp_ipi-uek7

Module with drgn api took 11 minutes with `smp_ipi-uek5`  vmcore, 5 minutes with `smp_ipi-uek6` and 6 minutes with `smp_ipi-uek7` whereas previous module took only 10 seconds, 4 seconds and 6 seconds with `smp_ipi-uek5`, `smp_ipi-uek6` and `smp_ipi-uek7` respectively.

**Comment by imran-kn** [comment]
> The updated module is using drgn api as suggested, but one thing we found is that this approach is slow compared to previous version (which was using partha's own api).
> 
> When ran on following gitlab test vmcores: smp_ipi-uek5, smp_ipi-uek6 and smp_ipi-uek7
> 
> Module with drgn api took 11 minutes with `smp_ipi-uek5` vmcore, 5 minutes with `smp_ipi-uek6` and 6 minutes with `smp_ipi-uek7` whereas previous module took only 10 seconds, 4 seconds and 6 seconds with `smp_ipi-uek5`, `smp_ipi-uek6` and `smp_ipi-uek7` respectively.

@srivathsa729 could you please confirm which of the helpers got slowed ?

**Comment by srivathsa729** [comment]
```
147 slub_helper = _get_slab_cache_helper(slab_cache)
148 slub_helper._slub_get_freelist(cpu_freelist, cpu_free_set)

imran-kn (approved):
LGTM

drgn_tools/slabinfo.py
brenns10: I'm revisiting the open issues on this PR just to make sure everything is handled. However, on a second look, this doesn't really bother me too much, so I'll just resolve the issue :)

drgn_tools/slabinfo.py
brenns10: This comment did not make sense, the code is correct. My apologies.

brenns10 (approved):
There are a few nitpicks I have with the type annotations and docstrings. However, I think this should be merged as-is, and I will fix up those nits afterwards. This is really excellent work, thanks!

PR #52, opened 2023-12-20
dm: add helper to dump device mapper devices

biger410:
The output is similar with 'dmshow' from crash-pykdump

```
>>> dm.show_dm(prog)
NUMBER  NAME                        MAPPED_DEVICE       FLAGS
dm-12   VGExaDb-LVDbOra1-real       0xffff940885630000  0x0
dm-6    VGExaDb-LVDbhome1           0xffff9551c7d30000  0x0
dm-11   VGExaDb-LVDoNotRemoveOrUse  0xffff9551c7ed0000  0x0
dm-7    VGExaDb-LVDblocal1          0xffff9551c7d80000  0x0
dm-1    VGExaDb-LVDbSys2            0xffff9551c7c40000  0x0
dm-8    VGExaDb-LVDbase1            0xffff9551c7df0000  0x0
dm-2    VGExaDb-LVDbSwap1           0xffff9551c7ca0000  0x0
dm-0    VGExaDb-LVDbSys1            0xffff9551c7c10000  0x0
dm-14   VGExaDb-LVDbOra1_snap       0xffff94fb16ca0000  0x0
dm-13   VGExaDb-LVDbOra1_snap-cow   0xffff941c1fa40000  0x0
dm-4    VGExaDb-LVDbgrid1           0xffff9551c7ce0000  0x0
dm-5    VGExaDb-LVDbmc1             0xffff9551c7d00000  0x0
dm-9    VGExaDb-LVDproduct1         0xffff9551c7e90000  0x0
dm-10   VGExaDb-LVDzabbix1          0xffff9551c7ea0000  0x0
dm-3    VGExaDb-LVDbOra1            0xffff9551c7cb0000  BLOCK_IO_FOR_SUSPEND|SUSPENDED|FROZEN
```

Gitlab ci test passed for this pr.

### DIFF
--- a/drgn_tools/dm.py
+++ b/drgn_tools/dm.py
@@ -1,0 +1,114 @@
+# Copyright (c) 2023, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
+"""
+Helpers for device mapper devices.
+"""
+import argparse
+from typing import Iterable
+from typing import Tuple
+
+from drgn import Object
+from drgn import Program
+from drgn.helpers.linux.list import list_for_each_entry
+from drgn.helpers.linux.rbtree import rbtree_inorder_for_each_entry
+
+from drgn_tools.corelens import CorelensModule
+from drgn_tools.module import ensure_debuginfo
+from drgn_tools.table import print_table
+from drgn_tools.util import BitNumberFlags
+from drgn_tools.util import kernel_version
+
+
+def for_each_dm_hash(prog: Program) -> Iterable[Tuple[Object, str]]:
+    for head in prog["_name_buckets"]:
+        for hc in list_for_each_entry(
+            "struct hash_cell", head.address_of_(), "name_list"
+        ):
+            yield hc.md, hc.name.string_().decode()
+
+
+def for_each_dm_rbtree(prog: Program) -> Iterable[Tuple[Object, str]]:
+    for hc in rbtree_inorder_for_each_entry(
+        "struct hash_cell", prog["name_rb_tree"], "name_node"
+    ):
+        yield hc.md, hc.name.string_().decode()
+
+
+def for_each_dm(prog: Program) -> Iterable[Tuple[Object, str]]:
+    if "_name_buckets" in prog:
+        return for_each_dm_hash(prog)
+    elif "name_rb_tree" in prog:
+        return for_each_dm_rbtree(prog)
+    else:
+        raise NotImplementedError("Cannot find dm devices")
+
+
+class DmFlagsBits(BitNumberFlags):
+    """
+    Class to convert preprocessor definitions to enum
+
+    drgn can't get the value of preprocessor definitions.
+    This is only appliable to the kernel starting 8ae126660fdd
+    which was merged by v4.10
+    """
+
+    BLOCK_IO_FOR_SUSPEND = 0
+    SUSPENDED = 1
+    FROZEN = 2
+    FREEING = 3
+    DELETING = 4
+    NOFLUSH_SUSPENDING = 5
+    DEFERRED_REMOVE = 6
+    SUSPENDED_INTERNALLY = 7
+    POST_SUSPENDING = 8
+    EMULATE_ZONE_APPEND = 9
+
+
+class DmFlagsBitsOld(BitNumberFlags):
+    """only appliable to kernel older than v4.10"""
+
+    BLOCK_IO_FOR_SUSPEND = 0
+    SUSPENDED = 1
+    FROZEN = 2
+    FREEING = 3
+    DELETING = 4
+    NOFLUSH_SUSPENDING = 5
+    MERGE_IS_OPTIONAL = 6
+    DEFERRED_REMOVE = 7
+    SUSPENDED_INTERNALLY = 8
+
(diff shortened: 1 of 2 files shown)


drgn_tools/util.py
```diff
@@ -113,6 +113,21 @@ def type_exists(prog: Program, type_: str) -> bool:
...
+    :param variable: variable name
+    :returns: true if exist, otherwise false
+    """
+    try:
+        prog.variable(variable)
+        return True
+    except LookupError:
+        return False
+
+
```
brenns10, line 130: Looking at the drgn-tools code, the convention for this seems to be `if "varname" in prog:`. Your function is more technically correct (as `"varname" in prog` would also check for functions and constants). But I think it may be more consistent to stick with the way the rest of the code does it.

drgn_tools/dm.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def for_each_dm_hash(prog: Program) -> Iterable[Tuple[Object, str]]:
+    # bucket size is 64
+    for i in range(64):
+        head = prog["_name_buckets"][i]
+        for hc in list_for_each_entry(
+            "struct hash_cell", head.address_of_(), "name_list"
+        ):
+            yield hc.md, hc.name.string_().decode()
```
brenns10, line 30: The declaration of `_name_buckets` is an explicitly-sized array:
```c
static struct list_head _name_buckets[NUM_BUCKETS];
```

So drgn already knows the size of the array and you can just iterate over it. No need to hardcode the `NUM_BUCKETS` value.

```suggestion
    for head in prog["_name_buckets"]:
        for hc in list_for_each_entry(
            "struct hash_cell", head.address_of_(), "name_list"
        ):
            yield hc.md, hc.name.string_().decode()
```

drgn_tools/dm.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def for_each_dm(prog: Program) -> Iterable[Tuple[Object, str]]:
+    if variable_exists(prog, "_name_buckets"):
+        dm_list = list(for_each_dm_hash(prog))
+    elif variable_exists(prog, "name_rb_tree"):
+        dm_list = list(for_each_dm_rbtree(prog))
+
+    for dm, name in dm_list:
+        yield dm, name
```
brenns10, line 47: You're creating the full list here, and then returning an iterator over that list. It doesn't really make sense to collect everything using `list()` unless you plan on returning the list directly. I would drop the `list()` call and directly return the iterables:
```suggestion
def for_each_dm(prog: Program) -> Iterable[Tuple[Object, str]]:
    if "_name_buckets" in prog:
        return for_each_dm_hash(prog)
    elif "name_rb_tree" in prog:
        return for_each_dm_rbtree(prog)
    else:
        raise NotImplementedError("Cannot find dm devices")
```

But if you want to return a fully-collected list, you could do that too:

```python
def for_each_dm(prog: Program) -> List[Tuple[Object, str]]:
    if "_name_buckets" in prog:
        return list(for_each_dm_hash(prog))
    elif "name_rb_tree" in prog:
        return list(for_each_dm_rbtree(prog))
    else:
        raise NotImplementedError("Cannot find dm devices")
```

drgn_tools/dm.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+class dm_flags_bits(BitNumberFlags):
```
brenns10, line 50: For Python classes, the preferred style would be something like `DmFlagsBits`, could you rename this?

drgn_tools/dm.py
```diff
@@ -0,0 +1,118 @@
...
+    FREEING = 3
+    DELETING = 4
+    NOFLUSH_SUSPENDING = 5
+    DEFERRED_REMOVE = 6
+    SUSPENDED_INTERNALLY = 7
+    POST_SUSPENDING = 8
+    EMULATE_ZONE_APPEND = 9
+
+
+class dm_flags_bits_old(BitNumberFlags):
```
brenns10, line 71: Similarly, `DmFlagBitsOld`

drgn_tools/dm.py
```diff
@@ -0,0 +1,118 @@
...
+
+    BLOCK_IO_FOR_SUSPEND = 0
+    SUSPENDED = 1
+    FROZEN = 2
+    FREEING = 3
+    DELETING = 4
+    NOFLUSH_SUSPENDING = 5
+    DMF_MERGE_IS_OPTIONAL = 6
+    DMF_DEFERRED_REMOVE = 7
+    DMF_SUSPENDED_INTERNALLY = 8
```
brenns10, line 82: I'm guessing you left the prefix `DMF_` on these in order to distinguish these flag names from the new flags. If you want that distinction visible in the table, then that's fine.

But it's easy to refer to both versions of the flags when they have the same name, e.g. `DmFlagBitsOld.DEFERRED_REMOVE` vs `DmFlagBits.DEFERRED_REMOVE`. So if you don't care about the string output, I'd prefer for all of these constants to appear without the `DMF_` prefix.

brenns10 (changes requested):

biger410:
Good comments, all addressed as suggestion. Thanks for the review.

brenns10 (approved):

brenns10:
Great, thank you!

PR #53, opened 2023-12-21
Task new helpers

imran-kn:
These helpers were added and reviewed when we were using internal repo for drgn-tools, but migration happened before these could be merged.

### DIFF
--- a/drgn_tools/task.py
+++ b/drgn_tools/task.py
@@ -27,6 +27,9 @@ from drgn.helpers.linux.pid import find_task
 from drgn.helpers.linux.pid import for_each_task
 from drgn.helpers.linux.sched import cpu_curr
 from drgn.helpers.linux.sched import task_state_to_char
+from drgn.helpers.linux.slab import find_slab_cache
+from drgn.helpers.linux.slab import slab_cache_for_each_allocated_object
+from drgn.helpers.linux.slab import slab_cache_is_merged
 
 from drgn_tools.corelens import CorelensModule
 from drgn_tools.mm import totalram_pages
@@ -56,6 +59,7 @@ TASK_KILLABLE = TASK_WAKEKILL | TASK_UNINTERRUPTIBLE
 TASK_STOPPED = TASK_WAKEKILL | __TASK_STOPPED
 TASK_TRACED = TASK_WAKEKILL | __TASK_TRACED
 TASK_IDLE = TASK_UNINTERRUPTIBLE | TASK_NOLOAD
+
 
 def nanosecs_to_secs(nanosecs: int) -> float:
     """
@@ -336,6 +340,7 @@ def show_tasks_last_runtime(tasks: Iterable[Object]) -> None:
         last_arrival = format_nanosecond_duration(time_nanosec)
         rows.append([last_arrival, state, pid, hex(t.value_()), cpu, command])
     print_table(rows)
+
 
 def is_task_in_state(task: drgn.Object, state: int) -> bool:
     """
@@ -890,6 +895,7 @@ def count_idle_tasks(prog: drgn.Program) -> int:
     count = sum(1 for _ in for_each_idle_task(prog))
 
     return count
+
 
 def show_taskinfo(prog: Program, tasks: Iterable[Object]) -> None:
     """
--- a/drgn_tools/task.py
+++ b/drgn_tools/task.py
@@ -35,6 +35,27 @@ from drgn_tools.util import has_member
 
 ByteToKB = 1024
 
+# tsk->state values
+TASK_RUNNING = 0x0000
+TASK_INTERRUPTIBLE = 0x0001
+TASK_UNINTERRUPTIBLE = 0x0002
+__TASK_STOPPED = 0x0004
+__TASK_TRACED = 0x0008
+# Used in tsk->exit_state
+EXIT_DEAD = 0x0010
+EXIT_ZOMBIE = 0x0020
+# Used in tsk->state again
+TASK_PARKED = 0x0040
+TASK_DEAD = 0x0080
+TASK_WAKEKILL = 0x0100
+TASK_WAKING = 0x0200
+TASK_NOLOAD = 0x0400
+TASK_NEW = 0x0800
+
+TASK_KILLABLE = TASK_WAKEKILL | TASK_UNINTERRUPTIBLE
+TASK_STOPPED = TASK_WAKEKILL | __TASK_STOPPED
+TASK_TRACED = TASK_WAKEKILL | __TASK_TRACED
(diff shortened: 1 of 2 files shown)


drgn_tools/task.py
```diff
@@ -20,11 +20,36 @@
...
+TASK_DEAD = 0x0080
+TASK_WAKEKILL = 0x0100
+TASK_WAKING = 0x0200
+TASK_NOLOAD = 0x0400
+TASK_NEW = 0x0800
+
+TASK_KILLABLE = TASK_WAKEKILL | TASK_UNINTERRUPTIBLE
+TASK_STOPPED = TASK_WAKEKILL | __TASK_STOPPED
+TASK_TRACED = TASK_WAKEKILL | __TASK_TRACED
+TASK_IDLE = TASK_UNINTERRUPTIBLE | TASK_NOLOAD
```
brenns10, line 51: As far as I know, these bits aren't constant, and in fact differ in UEK4:

```
include/linux/sched.h
209  #define TASK_RUNNING		0
210  #define TASK_INTERRUPTIBLE	1
211  #define TASK_UNINTERRUPTIBLE	2
212  #define __TASK_STOPPED		4
213  #define __TASK_TRACED		8
214  /* in tsk->exit_state */
215  #define EXIT_DEAD		16
216  #define EXIT_ZOMBIE		32
217  #define EXIT_TRACE		(EXIT_ZOMBIE | EXIT_DEAD)
218  /* in tsk->state again */
219  #define TASK_DEAD		64
220  #define TASK_WAKEKILL		128
221  #define TASK_WAKING		256
222  #define TASK_PARKED		512
223  #define TASK_STATE_MAX		1024
```

In particular, `TASK_DEAD` is 0x40 rather than 0x80 on UEK4, `TASK_PARKED` is 0x200 rather than 0x40, etc.

Drgn uses `task_state_array` to get the one-letter task state code definitions. Unfortunately, that doesn't appear to include the codes that are larger than `0x80`...

I don't know what the exact solution here would be - if you have use cases for these helpers, it might help to see which ones you're using. If you mainly depend on the basic state characters, then you could rely on drgn's `task_state_to_char()` but if not, then it may be necessary to create a table of state values based on the kernel version.

drgn_tools/task.py
```diff
@@ -203,6 +228,561 @@ def show_tasks_last_runtime(tasks: Iterable[Object]) -> None:
     print_table(rows)
 
 
+def is_task_in_state(task: drgn.Object, state: int) -> bool:
```
brenns10, line 231: My immediate thought about this helper is that it would be much more straightforward to have `task_state()` as a helper, to handle the difference in struct member name. Then callers could write their own if-statements such as: `if task_state(tsk) == TASK_RUNNING`.

Having dedicated `is_task_STATE()` functions adds a lot of mostly rote, duplicate code. I do get that it's a bit more ergonomic to be able to just import one function and use it directly, but I'm not sure if the ergonomics justifies the extra code. The same goes for the `for_each_STATE_task()` and `count_STATE_tasks()`. 

I think it would be good to have `task_state()`, `for_each_task_in_state()`, and `count_tasks_in_state()`, and ditch the rest of them. But I'm open to discussion on it!

brenns10 (changes requested):

drgn_tools/task.py
```diff
@@ -203,6 +228,561 @@ def show_tasks_last_runtime(tasks: Iterable[Object]) -> None:
...
+    """
+    Check if a task is in ``TASK_KILLABLE`` state.
+
+    :param task: ``struct task_struct *``
+    :returns: True if task is in ``TASK_KILLABLE`` state, False otherwise.
+    """
+    return is_task_in_state(task, TASK_KILLABLE)
+
+
+def is_task_struct_leaking(prog: drgn.Program) -> bool:
```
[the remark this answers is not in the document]
imran-kn: Thanks !

drgn_tools/task.py
imran-kn: The values are different only in UEK-4 and can be differentiated based on length of task_state_array which was of length 7 before v4.14 (so in UEK-4 as well). Also the differing states are PARKED, DEAD, WAKEKILL and WAKING only. I am unable to figure out how to declare the 2 set of values for these states. If I do something like 
`if len(prog["task_state_array"]) > 7:
            TASK_PARKED = 0x0040
            TASK_DEAD = 0x0080
            TASK_WAKEKILL = 0x0100
            TASK_WAKING = 0x0200
        else:
            TASK_PARKED = 0x0200
            TASK_DEAD = 0x0040
            TASK_WAKEKILL = 0x0080
            TASK_WAKING = 0x0100
`
Obviously, I get undefined name for prog. I am not sure what is the correct way to deal with such situations. Thoughts ?

I need all of these states,  drgn's task_state_to_run is good for usual cases but I intend to track tasks in other states as well because this info in useful in debugging issues involving wakeup_path and scheduler's load calculation/balancing logic.

brenns10: I should have replied to this sooner, please excuse my delay!

Yeah, at the global scope we can't use prog to conditionally define these things. So I think we would be stuck with something like this:

```python
def task_state_constants(prog: Program) -> Dict[str, int]:
    if len(prog["task_state_array"]) > 7:
        return {...}
    else:
        return {...}
```

If it's used frequently, you could definitely throw it into the `Program.cache` object.

imran-kn: > I should have replied to this sooner, please excuse my delay!
> 
> Yeah, at the global scope we can't use prog to conditionally define these things. So I think we would be stuck with something like this:
> 
> ```python
> def task_state_constants(prog: Program) -> Dict[str, int]:
>     if len(prog["task_state_array"]) > 7:
>         return {...}
>     else:
>         return {...}
> ```
> 
> If it's used frequently, you could definitely throw it into the `Program.cache` object.

@brenns10 I have addressed review comments and  asked for review again. I have removed helper is_task_struct_leaking because it needs some debugging on UEK-5. I created and used this helper while debugging a task_struct leak in UEK-6 but looks like for UEK-5 some more changes are needed. Anyways we can go ahead with helpers in this PR (as they are independent of removed helper). I will create a separate PR for that once I have debugged it.

drgn_tools/task.py
```diff
@@ -36,6 +36,105 @@
...
+
+
+def task_state_constants(prog: Program) -> Dict[str, int]:
+    task_states = {}
+
+    if len(prog["task_state_array"]) > 7:
+        try:
+            task_states = prog.cache["task_states_new"]
+        except KeyError:
+            task_states = {**task_states_common, **task_states_new}
```
brenns10, line 97: nice, I did not know that syntax worked. It even works on Python 3.6 :D

brenns10 (approved):
Looks great!

PR #54, opened 2023-12-21
Fixed logic for cpuinfo

Hannibal404:


brenns10:
This fixes the error we encountered, so I'll merge it. However, on UEK5, with the CTF debuginfo, the error did not occur, because `"cr4_read_shadow"` is not found in prog with CTF debuginfo, while it is found with DWARF debuginfo. I think the issue here is that there's no symbol named cr4_read_shadow, there's just a DWARF function entry for it.

Can we find a different detection method for that? We can keep it separate from this PR.

brenns10 (approved):

PR #55, opened 2023-12-21
Fixed bt crashing on running kernel

Hannibal404:


drgn_tools/bt.py
```diff
@@ -453,17 +454,22 @@ def print_all_bt(prog: Program) -> None:
...
+        for task in online_tasks:
+            bt(task)
+    else:
+        print("Cannot get stack of online tasks for running kernel")
 
     print("\nOff-CPU Tasks:")
     for task in for_each_task(prog):
-        if task.pid.value_() not in online_tasks:
+        if task not in online_tasks:
             bt(task)
```
brenns10, line 473: Unfortunately we can't rely on this method in general. On a live system, the set of online tasks is constantly changing. There's no guarantee that the tasks in `online_tasks` will actually still be running by the time they're printed. That also means that there's still a decent chance that some _other_ task will be running (and not in `online_tasks`) by the time we start printing the rest of the tasks out.

To avoid this, let's just catch the `ValueError` we get while unwinding the running tasks. We can reorganize the code so that we still show on-cpu tasks for vmcores.

```python
def print_all_bt(prog: Program) -> None:
    """
    Prints the stack trace of all tasks
    """
    online_tasks = set()

    if not prog.flags & ProgramFlags.IS_LIVE:
        print("On-CPU Tasks:")

        for cpu in for_each_online_cpu(prog):
            task = cpu_curr(prog, cpu)
            online_tasks.add(task.pid.value_())
            bt(task)
            print()

        print("\nOff-CPU Tasks:")

    for task in for_each_task(prog):
        if task.pid.value_() not in online_tasks:
            try:
                bt(task)
            except ValueError as e:
                # Catch ValueError for unwinding running tasks on live
                # systems. Print the task & comm but don't unwind.
                pid = task.pid.value_()
                comm = escape_ascii_string(task.comm.string_())
                print(f"PID: {pid}  COMM: {comm}\nerror: {str(e)}")
            print()
```

brenns10 (changes requested):

brenns10 (approved):

PR #57, opened 2024-01-08
Updates for block/nvme to run on RHCK & newer kernels

brenns10:
RHCK 9 contains entire subsystem backports from 5.18+ in the block & nvme. We'll probably need these for UEK8 too, but I wanted these so corelens would run on my laptop kernel. This is still a draft, I'm not sure that tests will pass for NVME.


brenns10:
Works on my laptop and on the OL vms which contain NVME drives. Tests pass on all vmcores.

biger410 (approved):

PR #64, opened 2024-01-23
Add crash net module

pkannoju:
Authored By: Praveen Kumar Kannoju<praveen.kannoju@oracle.com>
Co-Authored By: Srikanth C S<srikanth.c.s@oracle.com>
Co-Authored By: Shaleen Bathla<shaleen.bathla@oracle.com>

### DIFF
--- a/drgn_tools/crash_net.py
+++ b/drgn_tools/crash_net.py
@@ -1,0 +1,423 @@
+# Copyright (c) 2023, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
+import argparse
+import ipaddress
+import socket as socketlib
+from typing import Iterator
+from typing import List
+from typing import Tuple
+
+from drgn import Object
+from drgn import PlatformFlags
+from drgn import Program
+from drgn.helpers.linux import net as network
+from drgn.helpers.linux.fs import for_each_file
+from drgn.helpers.linux.list import hlist_for_each_entry
+from drgn.helpers.linux.list import list_for_each_entry
+from drgn.helpers.linux.net import SOCKET_I
+from drgn.helpers.linux.pid import find_task
+
+from drgn_tools.corelens import CorelensModule
+from drgn_tools.table import print_table
+
+_NETDEV_HASHBITS = 8
+_NETDEV_HASHENTRIES = 1 << _NETDEV_HASHBITS
+
+sockfamily_dict = {
+    0: "AF_UNSPEC",
+    1: "AF_UNIX",
+    2: "AF_INET",
+    10: "AF_INET6",
+    16: "AF_NETLINK/ROUTE",
+}
+
+socktype_dict = {
+    1: "SOCK_STREAM",
+    2: "SOCK_DGRAM",
+    3: "SOCK_RAW",
+    4: "SOCK_RDM",
+    5: "SOCK_SEQPACKET",
+    6: "SOCK_DCCP",
+    10: "SOCK_PACKET",
+}
+
+hwtype_dict = {
+    0: "ARPHRD_NETROM",
+    1: "ARPHRD_ETHER",
+    2: "ARPHRD_EETHER",
+    3: "ARPHRD_AX25",
+    4: "ARPHRD_PRONET",
+    5: "ARPHRD_CHAOS",
+    6: "ARPHRD_IEEE802",
+    7: "ARPHRD_ARCNET",
+    8: "ARPHRD_APPLETLK",
+    15: "ARPHRD_DLCI",
+    19: "ARPHRD_ATM",
+    23: "ARPHRD_METRICOM",
+    24: "ARPHRD_IEEE1394",
+    27: "ARPHRD_EUI64",
+    32: "ARPHRD_INFINIBAND",
+}
+
+state_dict = {
+    1: "NUD_INCOMPLETE",
+    2: "NUD_REACHABLE",
+    4: "NUD_STALE",
+    8: "NUD_DELAY",
+    16: "NUD_PROBE",
+    32: "NUD_FAILED",
+    64: "NUD_NOARP",
+    128: "NUD_PERMANENT",
+}
+
+
+def netdev_ipv4s(dev: Object) -> List[ipaddress.IPv4Address]:
+    """
+    Return a list of IPv4 addresses associated with a net_device.
+
+    :param dev: ``struct net_device *``
+    :returns: list of IPv4 addresses
+    """
+    ips = []
+    prog = dev.prog_
+    if dev.ip_ptr:
+        ifa = dev.ip_ptr.ifa_list
+        while ifa:
+            # ifa_address is already in network byte order: don't allow drgn to
+            # use the platform endianness to mangle it
+            addr_bytes = prog.read(ifa.ifa_address.address_of_(), 4)
+            ips.append(ipaddress.IPv4Address(addr_bytes))
+            ifa = ifa.ifa_next
+    return ips
+
+
+def netdev_ipv6s(dev: Object) -> List[ipaddress.IPv6Address]:
+    """
+    Return a list of IPv6 addresses associated with a net_device.
+
+    :param dev: ``struct net_device *``
+    :returns: list of IPv6 addresses
+    """
+    ips = []
+    prog = dev.prog_
+    if dev.ip6_ptr:
+        for addr in list_for_each_entry(
+            "struct inet6_ifaddr",
+            dev.ip6_ptr.addr_list.address_of_(),
+            "if_list",
+        ):
+            addr_bytes = prog.read(addr.addr.in6_u.u6_addr8.address_of_(), 16)
+            ips.append(ipaddress.IPv6Address(addr_bytes))
+    return ips
+
+
+def for_each_netdev(ns: Object) -> Iterator[Tuple[str, Object]]:
+    """
+    Iterate over each ``struct net_device *`` of a net namespace
+
+    :param ns: ``struct net *``
+    :returns: an iterator of ``struct net_device *``
+    """
+    # Since commit ff92741270bf ("net: introduce name_node struct
+    # to be used in hashlist"), net.dev_name_head is a list of name nodes. Prior
+    # to that, it contains the net_devices directly.
+    # The below approach is similar to what's already in drgn's
+    # netdev_get_by_name().
+    try:
+        entry_type = ns.prog_.type("struct netdev_name_node")
+        member = "hlist"
+        entry_is_name_node = True
+    except LookupError:
+        entry_type = ns.prog_.type("struct net_device")
+        member = "name_hlist"
+        entry_is_name_node = False
+
+    for i in range(_NETDEV_HASHENTRIES):
+        head = ns.dev_name_head[i]
+        for dev_node in hlist_for_each_entry(entry_type, head, member):
+            dev = dev_node.dev if entry_is_name_node else dev_node
+            yield (dev_node.name.string_().decode("utf-8"), dev)
+
+
+def print_net_devices(prog: Program) -> None:
+    """
+    Display the network devices, their names and addresses.
+    :param prog: drgn program
+    :returns: None.
+    """
+    rows = [["NET_DEVICE", "NAME", "IP ADDRESS(ES)"]]
+
+    net_namespaces = network.for_each_net(prog)
+    for net_namespace in net_namespaces:
+        for name, dev in for_each_netdev(net_namespace):
+            net_device_ptr = hex(dev.dev.platform_data)
+            ip_list = [str(a) for a in netdev_ipv4s(dev) + netdev_ipv6s(dev)]
+            ips = ", ".join(ip_list)
+            rows.append(
+                [
+                    net_device_ptr,
+                    name,
+                    ips,
+                ]
+            )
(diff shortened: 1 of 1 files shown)


drgn_tools/crash_net.py
```diff
@@ -0,0 +1,368 @@
...
+            )
+            print(socket)
+            print(socket.sk)
+        except Exception:
+            continue
+
+
+class CrashNet(CorelensModule):
+    """
+    Corelens Module for sysinfo
```
brenns10, line 335: `Displays details about network devices and sockets`

drgn_tools/crash_net.py
```diff
@@ -0,0 +1,368 @@
...
+    """
+
+    name = "net"
+
+    def add_args(self, parser: argparse.ArgumentParser) -> None:
+        parser.add_argument(
```
brenns10, line 341: Ok, so each of these arguments is mutually exclusive from one another. That is, `net -a -S PID` wouldn't make sense, right?

If that's the case, then you should use [add_mutually_exclusive_group()](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_mutually_exclusive_group), to enforce that.

```suggestion
    def add_args(self, parser: argparse.ArgumentParser) -> None:
        group = parser.add_mutually_exclusive_group()
        group.add_argument(
```

You use `group.add_argument()` for each of the arguments which are mutually exclusive to each other.

drgn_tools/crash_net.py
```diff
@@ -0,0 +1,368 @@
...
+            "-a",
+            dest="all_data",
+            action="store_true",
+            help="Displays net -a output",
```
brenns10, line 345:
```suggestion
            help="Displays the ARP table",
```
That way people don't need to know what crash's `net -a` does.

drgn_tools/crash_net.py
```diff
@@ -0,0 +1,368 @@
...
+        )
+        parser.add_argument(
+            "-s",
+            type=str,
+            dest="socket_data",
+            help="Displays net -s <PID> output",
```
brenns10, line 351: - Use `metavar="PID"` so that the help output will use PID to refer to the argument value.
- Use `type=int` to validate that a PID is an int, not a str.
- The help output shouldn't mention crash `net` command
```suggestion
            type=int,
            dest="socket_data",
            metavar="PID",
            help="Display a list of open sockets for PID",
```

drgn_tools/crash_net.py
```diff
@@ -0,0 +1,368 @@
...
+        )
+        parser.add_argument(
+            "-S",
+            type=str,
+            dest="socket_full_data",
+            help="Displays net -S <PID> output",
```
brenns10, line 357: As above:
```suggestion
            type=int,
            dest="socket_full_data",
            metavar="PID",
            help="Display open sockets for PID, with detailed data structure dumps",
```

drgn_tools/crash_net.py
```diff
@@ -0,0 +1,368 @@
...
+        )
+
+    def run(self, prog: Program, args: argparse.Namespace) -> None:
+        if args.all_data:
+            crash_a(prog)
+        elif args.socket_data:
+            crash_s(prog, args.socket_data)
+        elif args.socket_full_data:
+            crash_S(prog, args.socket_full_data)
```
brenns10, line 366: Like I mentioned in an earlier comment, you should be able to unify the two functions into something like `print_task_sockets()`, then you could do:
```suggestion
        elif args.socket_data:
            print_task_sockets(prog, args.socket_data, False)
        elif args.socket_full_data:
            print_task_sockets(prog, args.socket_full_data, True)
```

drgn_tools/crash_net.py
```diff
@@ -0,0 +1,368 @@
...
+state_dict = {
+    "1": "NUD_INCOMPLETE",
+    "2": "NUD_REACHABLE",
+    "4": "NUD_STALE",
+    "8": "NUD_DELAY",
+    "16": "NUD_PROBE",
+    "32": "NUD_FAILED",
+    "64": "NUD_NOARP",
+    "128": "NUD_PERMANENT",
+}
```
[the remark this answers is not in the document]
brenns10: Oh, as mentioned in a lot of comments below, you'd probably be well-served to simply remove the prefixes from all the comments. EG, instead of listing `AF_INET`, you could just list `INET`, and avoid some of the odd string manipulation you're currently doing to remove the prefix later.

drgn_tools/crash_net.py
pkannoju: Thank you, Stephen.
I will work on the review comments and send the updated version out for review.

drgn_tools/crash_net.py
```diff
@@ -0,0 +1,368 @@
...
+
+
+def net_crash(prog: Program) -> None:
```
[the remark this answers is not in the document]
brenns10: This should still be renamed. "net_crash()" doesn't reflect what this does, which is print network devices.

drgn_tools/crash_net.py
```diff
@@ -0,0 +1,368 @@
...
+                    ).reverse_pointer[:-13]
+                except KeyError:
+                    net_device_ipaddr = ""
+                print(
+                    "%-20s %-10s %-20s"
+                    % (net_device_ptr, net_device_name, net_device_ipaddr)
+                )
+
+
+def crash_a(prog: Program) -> None:
```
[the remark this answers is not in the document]
brenns10: This should still be renamed.

drgn_tools/crash_net.py
```diff
@@ -0,0 +1,368 @@
...
+                    destination_port,
+                )
+            )
+            print(socket)
+            print(socket.sk)
+        except Exception:
+            continue
+
+
+class CrashNet(CorelensModule):
```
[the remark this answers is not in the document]
brenns10: Please also rename this

drgn_tools/crash_net.py
brenns10: Please update this with a more useful description. This string is visible to users when they run `corelens -L`, in your case, the output would be like:
```
$ corelens -L
...
net  Corelens Module for sysinfo
```

Not only is this the wrong description, but it the user already knows they are running corelens and looking at its modules. Instead describe in a few words what the module does.

drgn_tools/crash_net.py
```diff
@@ -0,0 +1,372 @@
...
+            for j in range(4):
+                ip_addr += str(neigh.primary_key[j].value_()) + "."
+            ip_addr = ip_addr[:-1]
+
+            hw = neigh.ha
+            hw_addr = ""
+            for j in range(6):
+                hw_addr += str(hex(hw[j])[2:].zfill(2)) + ":"
+            hw_addr = hw_addr[:-1]
+            dev_name = neigh.dev.name.string_().decode()
```
brenns10, line 196: I would again suggest using drgn to directly read the bytes and then passing that to the ipaddress module here. Similarly for MAC addresses, it's a bit simpler to just use string formatting rather than using hex, str, and zfill manually.
```suggestion
            neigh_addr = hex(neigh.value_())[2:]
            ip_bytes = prog.read(neigh.primary_key.address_of_(), 4)
            ip_addr = str(ipaddress.IPv4Address(ip_bytes))
            
            mac_bytes = prog.read(neigh.ha.address_of_(), 6)
            hw_addr = ":".join(f"{b:02x}" for b in mac_bytes)
            dev_name = neigh.dev.name.string_().decode()
```

drgn_tools/crash_net.py
```diff
@@ -0,0 +1,372 @@
...
+
+    nht = prog["arp_tbl"].nht
+
+    hash_shift = nht.hash_shift.value_()
+    nhash_buckets = 1 << hash_shift
+    hash_buckets = nht.hash_buckets
+
+    for i in range(nhash_buckets):
+        neigh = hash_buckets[i]
+        while neigh:
```
brenns10, line 184: Re-iterating another comment from before:
It would be nice if you could factor out this part into a `for_each_neighbor(prog)` function!

drgn_tools/crash_net.py
```diff
@@ -0,0 +1,372 @@
...
+                        sock_family + ":" + sock_type,
+                        source_port,
+                        destination_port,
+                    )
+                )
+                print(socket)
+                print(socket.sk)
+
+        except Exception:
+            continue
```
brenns10, line 328: Copying over an unresolved comment from v1:

This `try/except` block spans 40 or 50 lines of code. I can't tell what error it's supposed to catch. And given that the `except` block catches `Exception`... well, anything could happen here and this function would just swallow errors and continue on.

In general, `try/except` blocks should either:
1. Be small, so that the reader can understand exactly where the expected error could come from, and why. Clearly, this try/except block is too big for that.
2. Serve some important error resilience purpose (for example: `corelens` catches all exceptions raised by a `CorelensModule` and logs them, but then continues to run the remaining modules -- that way corelens is resilient to logical errors in modules).

In this case, resilience isn't as important as correctness. If we can't properly format information about a socket, this function should crash, so that we can learn why and fix the function.

---

It seems to me that the main error you may care about is that the file is not corresponding to a socket. According to the [docs](https://drgn.readthedocs.io/en/latest/helpers.html#drgn.helpers.linux.net.SOCKET_I) for `SOCKET_I`, this raises a `ValueError` in that case. So if that's all you care about, you could do:

```python
try:
    socket = SOCKET_I(file)
except ValueError:
    # file is not a socket
    continue
```

drgn_tools/crash_net.py
```diff
@@ -0,0 +1,372 @@
...
+            "FD",
+            "SOCKET",
+            "SOCK",
+            "FAMILY:TYPE",
+            "SOURCE-PORT",
+            "DESTINATION-PORT",
+        ]
+    ]
+
+    for i in range(int(task.files.fdt.max_fds)):
```
brenns10, line 268: Drgn has a [helper](https://drgn.readthedocs.io/en/latest/helpers.html#drgn.helpers.linux.fs.for_each_file) for iterating over each open file for a task. Please use that instead.

drgn_tools/crash_net.py
```diff
@@ -0,0 +1,372 @@
...
+                int(socket.sk.__sk_common.skc_family)
+            ) not in sockfamily_dict.keys():
+                sock_family = "AF_"
+            sock_family = sock_family[3:]
+            sock_type = str(socktype_dict.get(int(socket.type)))
+            if int(socket.type) not in socktype_dict.keys():
+                sock_type = "SOCK_"
+            sock_type = sock_type[5:]
+
+            if sock_family == "INET":
```
brenns10, line 284: Please ensure we support IPv6 here as well.

drgn_tools/crash_net.py
```diff
@@ -0,0 +1,372 @@
...
+                sport = int(socket.sk.__sk_common.skc_num)
+                source_port = (
+                    str(ipaddress.ip_address(saddr).reverse_pointer[:-13])
+                    + "-"
+                    + str(sport)
+                )
+                destination_port = (
+                    ipaddress.ip_address(daddr).reverse_pointer[:-13]
+                    + "-"
+                    + str(dport)
```
brenns10, line 297: Again, the IP address reading is broken here, due to the same endianness concerns described in the previous revisions. And further, the ntohs() is only necessary on little endian platforms.

If it helps, you could make a little helper that takes a pointer to an IPv4 address and returns the Python IPv4Address data structure for it:

```python
def get_ipv4(ptr: Object) -> ipaddress.IPv4Address:
    return ipaddress.IPv4Address(prog.read_bytes(ptr, 4))

skc = socket.sk.__sk_common
saddr = get_ipv4(skc.skc_rcv_saddr.address_of_())
daddr = get_ipv4(skc.skc_daddr.address_of_())
sport = int(skc.skc_num)
dport = int(sokcet.skc_dport)
if prog.platform.flags & PlatformFlags.IS_LITTLE_ENDIAN:
    dport = socketlib.ntohs(dport)

source_port = f"{saddr}:{sport}"
destination_port = f"{daddr}:{dport}"
```

drgn_tools/crash_net.py
```diff
@@ -0,0 +1,372 @@
...
+    addresses and corresponding ports along with the dump of data
+    structures. If print_full_data is true, it prints detailed data
+    structure dumps.
+    :param prog: drgn program
+    :param pid: pid of the process.
+    :returns: None.
+    """
+
+    task = find_task(prog, int(pid))
+    if not (task):
```
brenns10, line 229: nit: no need for parantheses - `if not task:` is the standard way of expressing this

brenns10 (changes requested):
Thanks for this iteration. There are a few comments from my previous review that were overlooked. I commented on them if they were still visible on this diff, otherwise, I resolved them and re-posted them on this revision.

I do still think it would simplify your code significantly if you would change each of the dictionaries (`sockfamily_dict`) so that they don't contain the prefixes (e.g. AF_, SOCK_, etc.). But that's not a very high priority issue.

brenns10 (approved):
Thanks for bearing with my review. This is looking good from a Drgn perspective and basic networking perspective.

I will leave this as approved for the moment, we can verify whether we need a domain-specific review prior to merge.

pkannoju:
Remote node address in the arp output was being shown incomplete. Have fixed it and pushed the changes again.

mbpatil369:
I reviewed and approve this corelense net module. I am from net domain.

PR #65, opened 2024-01-24
Documentation updates, type annotation fixes, and new Table helpers

brenns10:
The partition info and slabinfo helpers had a couple small items that I wanted to address related to the docstrings and type annotations. I tweaked them and ensured that documentation builds successfully.

I also added two new table helpers: `Table` and `FixedTable`. Their docstrings should be self-explanatory, but the nice thing about them is that they can take care of the string formatting on your behalf, and they can also align individual columns to the left and right. By using `FixedTable`, you can also get output that is printed immediately, rather than all at once at the end of a loop. This is a major improvement for users, as it stinks to run a command and wait for a long time for a table to be output. To show the improvement, I moved the dentrycache helper to it, and the slabinfo helper, which can run a bit slow on heavily fragmented machines.

### DIFF
--- a/drgn_tools/table.py
+++ b/drgn_tools/table.py
@@ -3,29 +3,9 @@
 import sys
 from typing import Any
 from typing import Dict
+from typing import Iterable
 from typing import List
 from typing import Optional
-
-
-def print_row(fields: List[Any], col_widths: List[int]):
-    """
-    Print a single row of a table, given pre-determined column widths
-
-    Note that this doesn't guarantee that the width of every field in the row
-    will be less than the width of the column: in that case, the field's full
-    contents will be printed and columns will be misaligned. For guaranteed
-    aligned columns, see print_table(), or be very careful about your column
-    widths.
-
-    :param fields: a list of fields to print
-    :param col_widths: the width of each field (not including spaces)
-    """
-    print(
-        "  ".join(
-            str(val) if w <= 0 else str(val).ljust(w)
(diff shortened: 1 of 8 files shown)


brenns10:
Additional changes added here:

1. Previously the vmcore test logs only got uploaded to gitlab when the test succeeded. Obviously that's less than ideal, so I've fixed that.
2. The recently merged slabinfo helper is great, but doesn't support the recent 5.17 change from `struct page` to `struct slab`. I've added support for that here, which helps support UEK8 and RHCK on OL9.

drgn_tools/table.py
```diff
@@ -66,6 +46,174 @@ def print_table(
...
+            print(self._row_str(row), file=self.out)
+
+
+class FixedTable(Table):
+    """
+    Created an aligned, formatted table with fixed column widths
+
+    This is a variant of the :class:`Table` class, and it is designed to be
+    nearly a drop-in replacement. Unlike :class:`Table`, the column widths are
+    fixed  Unlike :class:`Table`, the column widths are "fixed"; that is they
```
biger410, line 158: There is a duplicate of "Unlike :class:`Table`, the column widths are "fixed""

biger410 (approved):
one small comment, you may fix it when merging. other looks good.

brenns10:
Thanks @biger410, unfortunately I can't merge directly so I'll need you to re-approve.

biger410 (approved):

PR #66, opened 2024-01-31
Fix bt and release 0.8.0

brenns10:
Updates `bt` to skip zombie tasks, and adds some filtering logic for the `bt` corelens module which I think will be quite useful.

Relase 0.8.0 after that, for packaging up to our various environments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,32 @@ Unreleased
 
 Changes which are committed to git, but not yet released, may appear here.
 
+0.8.0 - Wed, Jan 31, 2023
+-------------------------
+
+Another intermediate release prior to 1.0.0. We have a few new corelens modules,
+and some fixes and general improvements.
+
+### Added
+
+- `slabinfo` corelens module
+- `partitioninfo` corelens module
+- `CtfCompatibility` to automatically detect which UEK kernels are compatible
+  with CTF debuginfo
+
+### Fixed
+
+- `bt` corelens module is more resilient to unwind errors
+- `md` helpers now support kernels containing a92ce0feffee ("md: drop queue
+  limitation for RAID1 and RAID10"), from 5.17
+
+### Changed
+
+- `bt()` function now outputs the task state in `[]` brackets as part of the
+  header
+- `scsiinfo` now outputs more useful fields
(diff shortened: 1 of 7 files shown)


biger410 (commented):
I thought dump all tasks will be very helpful even though it may take time, at least the on cpu task call trace can help us know easily spin lock spinning issue that we had a lot of customer cases.

brenns10:
> I thought dump all tasks will be very helpful even though it may take time, at least the on cpu task call trace can help us know easily spin lock spinning issue that we had a lot of customer cases.

We can still have customers run `corelens CORE -M bt`, this just means that the default `corelens CORE -a` report will not contain _every task's_ output.

biger410:
> > I thought dump all tasks will be very helpful even though it may take time, at least the on cpu task call trace can help us know easily spin lock spinning issue that we had a lot of customer cases.
> 
> We can still have customers run `corelens CORE -M bt`, this just means that the default `corelens CORE -a` report will not contain _every task's_ output.

I think if some information is good help to troubleshooting issues, we should put it into corelens default except it is extremely slow or have other limitations. Then once we generate a corelens report, we can go through each one to see whether is any obvious issues. I fully understand dump call trace of all tasks may take time and most of the call trace may be not userful, how about we dump call trace only for D and on-cpu processes?

brenns10:
I think I agree. However, I really didn't like the idea of having a `bt` command that does something weird like print on-CPU tasks followed by D-state tasks. It doesn't make intuitive sense to have a command which does that.

I decided to make the `bt` command behave somewhat similarly to crash:

* `bt PID | task` - unwinds a single task
* `bt -c CPU` - unwinds the task on a CPU
* `bt -a` - unwinds all on-CPU tasks
* `bt -A` - unwinds every task on the system
* `bt -s D` - unwinds every task in a given state (e.g. D)

I made the default behavior of the corelens report to run:

    bt -a
    bt -s D

I needed to tweak corelens a bit, so that it can run modules multiple times in a report. But with that, I think this is very useful for customer corelens reports, as well as for our own usage.

biger410:
This looks great to me.

biger410 (approved):

PR #67, opened 2024-02-01
task: add more options to "ps" module

SiddhiK29:
Allow passing pid | task | command to filter output by these. 
-k: display only kernel threads
-u: display only user threads
-G: display only the thread group leader

### DIFF
--- a/tests/test_task.py
+++ b/tests/test_task.py
@@ -1,8 +1,10 @@
 # Copyright (c) 2023, Oracle and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
+from drgn.helpers.linux.pid import find_task
 from drgn.helpers.linux.pid import for_each_task
 
 from drgn_tools import task
+from drgn_tools.workqueue import for_each_worker
 
 
 def test_show_taskinfo(prog):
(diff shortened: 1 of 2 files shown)


tests/test_task.py
```diff
@@ -10,3 +11,37 @@ def test_show_taskinfo(prog):
     task.show_tasks_last_runtime(for_each_task(prog))
     print("===== Display task information =====")
     task.show_taskinfo(prog, for_each_task(prog))
+
+
+def test_show_argwise_taskinfo(prog):
```
brenns10, line 16: I really appreciate adding testing for this!

This test, like most of the others we have, is a smoke test: it just executes some code (which prints out data) and does not do any checks for the validity or correctness of the result. Smoke tests are very useful for verifying that our helpers can at least _run_ on different kernel versions, so we can detect relevant code changes.

That said, the current `test_show_taskinfo()` already does 99% of the smoke testing we really need here. In this case, we're simply printing every task on the system several times over. The only additional code we're executing is the `is_user`, `is_kthread`, and `is_group_leader` tests. Since printing lots of data (some vmcores are from very large systems with many tasks) isn't free, this has the potential to slow down our test suite a lot, without really adding much value.

Instead, I'd really appreciate a simpler test that checks the logic of these three functions directly. For instance:
- We know that the init task will always be a user task, and a group leader.
- The threads of init are also user tasks, but they aren't group leaders.
- The workqueue system maintains sets of kthread workers, these are kernel threads.

We could write a quite efficient test like this:

```python
from drgn_tools.workqueue import for_each_worker


def test_user_kernel_threads(prog):
    init = find_task(prog, 1)
    assert task.is_user(init)
    assert task.is_group_leader(init)

    kworker = next(for_each_worker(prog)).task
    assert task.is_kthread(kworker)
    assert not task.is_user(kworker)
    assert task.is_group_leader(kworker)
```

This is quite quick as it just looks at tasks we know are there (init, and kworkers), and it doesn't print anything, nor does it try to iterate over every task.

brenns10 (changes requested):
One comment regarding testing, and one regarding a slight reorganization that may help reduce the code duplication. Sorry for the delayed review!

brenns10 (approved):

PR #72, opened 2024-02-22
Some bt performance improvements

brenns10:
Most of this relates to `bt_has()` which is used by the locking helpers. I added two improvements:

1. Indexing function calls for vmcores. This means that after the first `bt_has()` call, subsequent calls are basically free, since they use the index.
2. Creating a `bt_has_any()` API to allow searching for multiple functions at a time. This improves performance for live systems (though `bt_has()` is inherently a bad fit for live systems)...
3. Remove a duplicate call to `stack_trace()` as this incurs extra overhead


biger410 (approved):
looks good.

PR #74, opened 2024-02-22
Add the list_lru interator helper

marktinguely:
list_lru iterators for UEK5, UEK6, UEK7 and UEK8.

The list_lru iteration can iterate all memcg and NUMA nodes or one memcg index and one NUMA node.

list_lru_kmem_to_memcgidx() is provided to calculate the memcg index for a list_lru kvm address. This routine is written for list_lru kernel address and assumes slab allocated addresses.

list_lru_kmem_to_nodeid() is provided to calculate the NUMA node id for a list_lru kvm address. 

tests/test_list_lru.py uses every function to walk filesystem(s) and gets the memcg index and NUMA node id of the dentry (and xfs_buf on XFS filesystems) and uses those values to verify that list_lru_kmem_to_memcgidx() finds the entry. 

There is an example that counts negative dentry at:
 https://github.com/marktinguely/examples/blob/main/count_neg_dentry.py

### DIFF
--- a/drgn_tools/list_lru.py
+++ b/drgn_tools/list_lru.py
@@ -1,0 +1,305 @@
+# Copyright (c) 2025, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
+"""
+LRU Lists
+------------
+
+Helper to work with LRU lists. LRU can be created to be memcg aware and
+ordered by NUMA node.
+
+The routines iterate through the specified LRU and on NUMA machines, the
+output keeps the entries ordered by NUMA node.
+
+The list_lru_for_each_list() function iterates all of the list_lru_one
+list. The list_lru_for_each_entry() function iterates through all the
+specified entries on a list_lru and returns the NUMA nodeid, memcg
+and Object of the specified type.
+
+The list_lru_from_memcg_node_for_each_list() and
+list_lru_from_memcg_node_for_each_entry() functions allows the user to
+restrict the iteration of the list_lru_one and entries by the memcg
+index when the list_lru is memcg_aware and the NUMA node identifier.
+
+slab_object_to_memcgidx() is a helper to find the mem_cgroup index
+from a list_lru entry. This routine is only interested in slab
+allocated entries and does not check nor handle the MEMCG_DATA_KMEM case.
+slab_object_to_nodeid() is a helper to find the NUMA node id from a
+list_lru entry.
+"""
+from typing import Iterator
+from typing import Tuple
+from typing import Union
+
+from drgn import cast
+from drgn import IntegerLike
+from drgn import NULL
+from drgn import Object
+from drgn import Type
+from drgn.helpers.linux.list import list_for_each_entry
+from drgn.helpers.linux.mm import compound_head
+from drgn.helpers.linux.mm import page_to_pfn
+from drgn.helpers.linux.mm import page_to_virt
+from drgn.helpers.linux.mm import virt_to_page
+from drgn.helpers.linux.nodemask import for_each_online_node
+from drgn.helpers.linux.nodemask import node_state
+from drgn.helpers.linux.slab import slab_object_info
+from drgn.helpers.linux.xarray import xa_for_each
+from drgn.helpers.linux.xarray import xa_load
+
+from drgn_tools.meminfo import get_active_numa_nodes
+from drgn_tools.util import has_member
+
+__all__ = (
+    "list_lru_for_each_list",
+    "list_lru_for_each_entry",
+    "list_lru_from_memcg_node_for_each_list",
+    "list_lru_from_memcg_node_for_each_entry",
+    "slab_object_to_memcgidx",
+    "slab_object_to_nodeid",
+)
+
+
+def list_lru_for_each_list(lru: Object) -> Iterator[Tuple[int, int, Object]]:
+    """
+    Iterate over a list_lru and return each NUMA nodeid, memcgid and
+    list_lru_one object.
+
+    :param lru: ``struct list_lru *``
+    :return: Iterator of the Tuple (node_id, memcg_idx, ``list_lru_one *``)
+    """
+    prog = lru.prog_
+    memcg_aware = 0
+    # v5.2-rc2-303-g3e8589963773 (memcg: make it work on sparse non-0-node
+    # systems) adds memcg_aware boolean
+    if has_member(lru, "memcg_aware") and lru.memcg_aware:
+        memcg_aware = 1
+
+    # Before v5.15.0-9.96.3-944-gd337fa4c0eb2 (Oracle) and
+    # v5.17-47-g6a6b7b77cc0f (community) (mm: list_lru: transpose the array
+    # of per-node per-memcg lru lists), the list_lru_memcg entry was in the
+    # list_lru_node that was in the list_lru.
+    if has_member(lru, "node"):
+        if has_member(lru.node, "memcg_lrus") and lru.node[0].memcg_lrus:
+            memcg_aware = 1
+
+    if memcg_aware:
+        if has_member(lru, "ext") or has_member(lru, "xa"):
+            if has_member(lru, "ext"):
+                # (uek7) Oracle port UEK_KABI_REPLACE of node to ext
+                # v5.15.0-9.96.3-944-gd337fa4c0eb2 of commity patch
+                # v5.17-47-g6a6b7b77cc0f
+                xa = lru.ext.xa
+            else:
+                # uek 8 v5.17-57-gbbca91cca9a9 replace array with xarray
+                # doesn't have uek7 KABI changes.
+                xa = lru.xa
+            # Keep the entries grouped by the NUMA node.
+            for nid in for_each_online_node(prog):
+                for memcgid, memcg in xa_for_each(xa.address_of_()):
+                    # convert from the void ptr
+                    memcg = Object(prog, "struct list_lru_memcg *", memcg)
+                    yield (nid, memcgid, memcg.node[nid])
+        else:
+            for nid in for_each_online_node(prog):
+                # Keep the entries grouped by the NUMA node.
+                for i in range(prog["memcg_nr_cache_ids"]):
+                    yield (nid, i, lru.node[nid].memcg_lrus.lru[i])
+    else:
+        # not lru.memcg_aware
+        for nid in for_each_online_node(prog):
+            if has_member(lru, "ext"):
+                yield (nid, 0, lru.ext.node[nid].lru)
+            else:
+                yield (nid, 0, lru.node[nid].lru)
+
+
+def list_lru_for_each_entry(
+    type: Union[str, Type], lru: Object, member: str
+) -> Iterator[Tuple[int, int, Object]]:
+    """
+    Iterate over all of the entries in a list_lru.
+    This function calls list_lru_for_each_list() and then iterates over
+    each list_lru_one.
+
+    :param type: Entry type.
+    :param lru: ``struct list_lru *``
+    :param member: Name of list node member in entry type.
+    :return: Iterator of ``type *`` objects.
+    """
+    for nid, memcgid, llru1 in list_lru_for_each_list(lru):
+        for entry in list_for_each_entry(
+            type, llru1.list.address_of_(), member
+        ):
+            yield (nid, memcgid, entry)
+
+
+def list_lru_from_memcg_node_for_each_list(
+    mindx: IntegerLike,
+    nid: IntegerLike,
+    lru: Object,
+) -> Object:
+    """
+    Iterate over each list_lru_one entries for the provided memcg and NUMA node.
+
+    :param mindx: memcg index.
+    :param nid: NUMA node ID.
+    :param lru: ``struct list_lru *``
+    :return: Iterator of ``struct list_lru_one`` objects.
+    """
+    prog = lru.prog_
+    if node_state(nid, prog["N_ONLINE"]):
+        memcg_aware = 0
+        # v5.2-rc2-303-g3e8589963773 (memcg: make it work on sparse non-0-node
+        # systems) adds memcg_aware boolean
+        if has_member(lru, "memcg_aware") and lru.memcg_aware:
+            memcg_aware = 1
+        # Before v5.15.0-9.96.3-944-gd337fa4c0eb2 (Oracle) and
+        # v5.17-47-g6a6b7b77cc0f (community) (mm: list_lru: transpose the array
+        # of per-node per-memcg lru lists), the list_lru_memcg entry was in the
+        # list_lru_node that was in the list_lru.
+        if has_member(lru, "node"):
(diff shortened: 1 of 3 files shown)


drgn_tools/list_lru.py
```diff
@@ -0,0 +1,204 @@
...
+    Iterate over all of the entries in a list_lru.
+
+    :param prog: Kernel being debugged
+    :param type: Entry type.
+    :param lru: ``struct list_lru *``
+    :param member: Name of list node member in entry type.
+    :return: Iterator of ``type *`` objects.
+    """
+    if ((has_member(lru, "memcg_aware") and lru.memcg_aware) or
+        (has_member(lru.node, "memcg_lrus") and lru.node[0].memcg_lrus)) :
```
marktinguely, line 54: sorry this has to be made smarter for uek7 memcg unaware lrus. 
I will  respin, test everywhere and make a new pull request

brenns10 (commented):
Since you're already doing a respin, I wanted to give a little bit of feedback that should reduce the round trips on this review. The code logically is looking great, but stylistically, there's some things that don't fit with the project's existing code style. I don't want to waste developer's time or mental effort on code style though, so we use some automatic code formatters & linters to help catch these things at commit time.

It's documented [here](https://oracle-samples.github.io/drgn-tools/code-quality.html), the checks happen via a pre-commit hook. You can set it up via:

```
# OL9
dnf install -y pre-commit

# OL8
dnf install python39-pip
python3.9 -m pip install --user pre-commit

# Then, for both:
cd path/to/drgn-tools
pre-commit install --install-hooks
```

Next time you commit (including amending a commit), the hook will check code. Some checks (e.g. code style) automatically make the changes to your code, so you'll just need to `git add` those changes. Other checks are just linters, so they may surface an issue you'll need to manually address.

drgn_tools/list_lru.py
```diff
@@ -0,0 +1,236 @@
...
+            # not lru.memcg_aware
+            if has_member(lru, "ext"):
+                li = lru.ext.node[nid].lru.list
+            else:
+                li = lru.node[nid].lru.list
+            for entry in list_for_each_entry(type, li.address_of_(), member):
+                yield entry
+
+
+def list_lru_kmem_to_memcgidx(prog: Program, kvm: IntegerLike) -> IntegerLike:
```
marktinguely, line 186: In theory, we could get the numa node id from the page ptr, BUT there are multiple ways that is can be encoded.

Option 1: it is encoded in the vm_page.flags with the offset determined by what else is embedded in the vm_page.flags (section, zone, last cupid)
Option 2: it is  in a special section_to_node_table[] array where the section is encoded in the the vm_page.flags

A check for the section_to_node_table[] is easy enough but the vm_page.flags offsets/width are CONFIG dependent and is less obvious for Drgn  find automatically. From disassembling the code, the hosts that I have, the numa node is:
   (page.flags  >>  0x36)

marktinguely:
I wrote a simple drgn script that uses list_lru to count the dentry and negative dentry for one or more fileystems using for_each_mount()

drgn_tools/list_lru.py
brenns10: I have something like this, which depends on the `node_data` variable (`get_active_numa_nodes()` uses it):
```python
def page_to_node(pg: Object) -> Object:
    pfn = page_to_pfn(pg)
    nodes = get_active_numa_nodes(pg.prog_)
    for i in range(1, len(nodes)):
        if nodes[i - 1].node_start_pfn <= pfn < nodes[i].node_start_pfn:
            return nodes[i - 1]
    return nodes[-1]
```
It's decidedly inelegant, and if necessary, it could be made more efficient by turning this into a tree or something. But the benefit is that it doesn't require any knowledge of the number of bits to shift in `page.flags`. I have not yet determined a reliable way to determine that number of bits, sadly.

drgn_tools/list_lru.py
```diff
@@ -0,0 +1,261 @@
...
+                return memcgrp.kmemcg_id
+        else:
+            return -1
+    else:
+        # Before v5.13
+        scache = cpage.slab_cache
+        if scache == NULL(prog, "struct kmem_cache *"):
+            return -1
+        else:
+            return cpage.slab_cache.memcg_params.memcg.kmemcg_id
```
brenns10, line 241: I believe you can rework much of this to use [slab_object_info()](https://drgn.readthedocs.io/en/latest/helpers.html#drgn.helpers.linux.slab.slab_object_info), which can tell you the slab_cache, starting address, and the `struct slab` or page associated with the object. That can help you avoid the compound head call, the conditional casts to struct slab, etc.

drgn_tools/list_lru.py
```diff
@@ -0,0 +1,261 @@
...
+    Return the NUMA node id of the list_lru entry.
+
+    :param prog: Kernel being debugged
+    :param kvm: address of a list_lru entry
+    :return: NUMA node id
+    """
+    page = virt_to_page(prog, kvm)
+    cpage = compound_head(page)
+    #
+    # Stephen Brennan routine to convert a page to node
```
brenns10, line 255: no need to include my name here. I hope I'll actually be able to come up with a better way, so I don't even necessarily want my name associated with this code 😜

brenns10 (changes requested):
Sorry for the inexcusable delay here. I picked these changes and played around with them in the dentry cache. I ended up having some improvements I'd like to see, but otherwise I'm excited to get these merged.

drgn_tools/list_lru.py
```diff
@@ -0,0 +1,261 @@
...
+
+    :param prog: Kernel being debugged
+    :param mindx: memcg index.
+    :param nid: NUMA node ID.
+    :param type: Entry type.
+    :param lru: ``struct list_lru *``
+    :param member: Name of list node member in entry type.
+    :return: Iterator of ``type *`` objects.
+    """
+    if node_state(nid, prog["N_ONLINE"]):
```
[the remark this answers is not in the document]
marktinguely: It was added because an invalid node id could be supplied.
since it is used as an index into memory, this is an attempt to sanity test the parameter.

drgn_tools/list_lru.py
marktinguely: I will look at it deeper but after a quick look, I  do not think it provides the information that I needed to get the memcg

marktinguely:
I have been debating for a while about changing the API returning the numa node id and memcg with the list_lru_for_each_entry(). The caller can easily ignore them with a "_, _, object = list_lru_for_each_entry(llru)". 

I still have to split list_lru_from_memcg_node_for_each_entry() into two routine and there is a lot of testing to redo. I ran it on LUCI for giggles and it works.

brenns10:
Please do ping me on Slack when you're ready for additional review & merge, that way I know when to check your update (and I don't delay too long). Thanks!

marktinguely:
> Please do ping me on Slack when you're ready for additional review & merge, that way I know when to check your update (and I don't delay too long). Thanks!

---
I dId the split of the foreach into list/entry.

I changed the API to return the memcg and NUMA node id back with the lru entry. Today, I decided to use this to do a quick (just verify the memcg and NUMA node O(n)) versus deep test (walk the memcg and NUMA node subtree to find the entry which is basically O(n^2)).

This has been tested on uek5 to Luci. I tried on a 1, 2 and 8 node NUMA. With a few to 5 million dentrys.

The slab code doesn't give me what I need to find the memcg. The memcg and NUMA node from kmem routines are really would be used in a test environment.

drgn_tools/list_lru.py
```diff
@@ -0,0 +1,282 @@
...
+                return memcgrp.kmemcg_id
+        else:
+            return -1
+    else:
+        # Before v5.13
+        scache = cpage.slab_cache
+        if scache == NULL(prog, "struct kmem_cache *"):
+            return -1
+        else:
+            return cpage.slab_cache.memcg_params.memcg.kmemcg_id
```
brenns10, line 263: One thing we really try to do is cite each commit, or at least one of the commits in the relevant patch series, along with the kernel version, in order to justify each change in behavior.

And again, I think that using the `slab_object_info()` helper can really clean this up. Here's how I would code up this helper:

```python
def slab_object_to_memcgidx(obj: Object):
    prog = obj.prog_
    info = slab_object_info(obj)
    if not info:
        raise ValueError("not a slab object")

    if hasattr(info.slab_cache, "memcg_params"):
        # Prior to v5.9, there were separate slab caches per memcg, so the memcg
        # could be determined from the slab cache itself.
        params = info.slab_cache.memcg_params
        if params.memcg:
            return params.memcg.kmemcg_id.value_()
        else:
            return -1

    slab_object_index = (
        obj.value_()
        - page_to_virt(info.slab).value_()
    ) // info.slab_cache.size.value_()

    if hasattr(info.slab, "obj_cgroups"):
        # Starting with v5.9 in commit 10befea91b61c ("mm: memcg/slab: use a
        # single set of kmem_caches for all allocations"), until v5.11, object
        # cgroup information was stored in a "obj_cgroups" array, which was
        # shared in a union as "mem_cgroup". The lowest bit is set to indicate
        # that it is an array of object cgroup information.
        obj_cgroups = info.slab.obj_cgroups
        if not obj_cgroups.value_() & 1:
            return -1
        memcg_data = Object(prog, obj_cgroups.type_, obj_cgroups.value_() - 1)
        memcg = memcg_data[slab_object_index].memcg
    elif hasattr(info.slab, "memcg_data"):
        # Starting with v5.11 commit 87944e2992bd2 ("mm: Introduce page memcg
        # flags"), until v6.10 , the "mem_cgroup" and "obj_cgroups" are placed
        # into the unsigned long field "memcg_data", with constant flags to
        # formalize the access to them.
        flag = prog.constant("MEMCG_DATA_OBJCGS")
        mask = cast("unsigned long", prog.constant("__NR_MEMCG_DATA_FLAGS") - 1)
        if not info.slab.memcg_data & flag:
            return -1
        memcg_data = cast("struct obj_cgroups **", info.slab.memcg_data & ~mask)
        memcg = memcg_data[slab_object_index].memcg
    elif hasattr(info.slab, "obj_exts"):
        # Since v6.10 commit 21c690a349baa ("mm: introduce slabobj_ext to
        # support slab object extensions"), struct slab now supports more type
        # of object metadata, in addition to memcg info. There are new constants
        # to check and a new type, slabobj_ext, to use for accessing the
        # metadata.
        flag = prog.constant("MEMCG_DATA_OBJEXTS")
        mask = cast("unsigned long", prog.constant("__NR_OBJEXTS_FLAGS") - 1)
        if not info.slab.obj_exts & flag:
            return -1
        exts = cast("struct slabobj_ext *", info.slab.obj_exts & ~mask)
        memcg = exts[slab_object_index].objcg.memcg
    else:
        raise RuntimeError(
            "Cannot find object memcg info for this kernel version"
        )

    if memcg:
        return memcg.kmemcg_id.value_()
    else:
        return -1
```

tests/test_list_lru.py
```diff
@@ -0,0 +1,157 @@
...
+    verify: Optional[IntegerLike] = None,
+) -> None:
+    """
+    Tests memcg aware and unaware lru by walking the entire lru and for
+    every entry found, look up the memcg index (may be -1) and NUMA node ID.
+    Compare the memcg and NUMA node id with the one returned by list_lru
+    iterator. When "verify" is provided, walk the specific portion of the
+    list_lru to verify this entry is found by memcg index and NUMA node ID.
+    This is an exhastive search of every entry.
+    """
```
brenns10, line 37: The test looks quite useful for manual testing, but I think there's a few improvements we could make for automated testing.

1. We're iterating over _every_ mount (which could have the same super block multiple times...) and then iterating over _every_ dentry in the LRU for the super block. Which is just a lot of iteration and dentries. I think instead, we should limit the number of dentries and super blocks. This way we can test out that the helper is handling the data structures, but we don't need to waste time running through all the (potentially millions) of elements in each CI run. We're only going to add more vmcores & test VMs to the suite, so making sure tests run quickly on each one is paramount.
2. Rather than counting errors and printing them in verbose mode, I'd prefer to see assertions used, so that the test actually fails on any unexpected values.

brenns10 (changes requested):

drgn_tools/list_lru.py
```diff
@@ -0,0 +1,282 @@
...
+                    memcg = Object(prog, "struct list_lru_memcg *", memcg)
+                    yield (nid, memcgid, memcg.node[nid])
+        else:
+            # Before v5.13, memcg entries are in an array
+            # Keep the entries grouped by the NUMA node.
+            for nid in for_each_online_node(prog):
+                for i in range(prog["memcg_nr_cache_ids"]):
+                    llru1 = lru.node[nid].memcg_lrus.lru[i]
+                    if not list_empty(llru1.list.address_of_()):
+                        yield (nid, i, llru1)
```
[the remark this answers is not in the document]
marktinguely: done. it can be pretty sparse. the API was changed several times because it was discovered that it wasted space

drgn_tools/list_lru.py
```diff
@@ -0,0 +1,282 @@
...
+            type, llru1.list.address_of_(), member
+        ):
+            yield (nid, memcgid, entry)
+
+
+def list_lru_from_memcg_node_for_each_list(
+    mindx: IntegerLike,
+    nid: IntegerLike,
+    lru: Object,
+) -> Iterator[Object]:
```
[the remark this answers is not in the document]
marktinguely: yes, it not an iterator

tests/test_list_lru.py
brenns10: Regarding the full search -- it's really a question of a trade-off between time and utility. The main utility of the tests is to detect when the kernel types have changed and the helpers don't work anymore. At least 80-90% of that goal can be achieved with a test that just runs through a few members of a memcg-aware LRU, and a non-aware LRU, and validates that they can be found through the respective APIs. The remaining 10-20% may require exhaustively looking at each element of the LRU, but given that dentry LRUs can contain millions of elements, that could seriously slow down the tests, for a marginal gain.

That said, I'll take a look at the runtimes on the various vmcores we have and see what the runtime is before I make any conclusions :)

drgn_tools/list_lru.py
marktinguely: fails on uek7 v5.15.0-300.163.18.7.el8uek.x86_64

brenns10: My apologies, the structure name `struct obj_cgroups` should have been `struct obj_cgroup`. I've updated the code.

marktinguely: uek6 works with the change:
    if hasattr(info.slab_cache, "memcg_params"):
        # Prior to v5.9, there were separate slab caches per memcg, so the memcg
        # could be determined from the slab cache itself.
        params = info.slab_cache.memcg_params
        if params.memcg:
            return params.memcg.kmemcg_id.value_()
-        else:             <<< remove
-            return -1   <<< remove

tests/test_list_lru.py
marktinguely: added an optional "maxitems" parameter to limit the search through a particular filesystem's dentry/xfs_buf entries.

marktinguely:
thanks you for all the great ideas and help

PR #75, opened 2024-02-26
Add (Gitlab CI) testing with CTF enabled, plus improved parallelism

brenns10:
This branch adds lots of improvements to the Gitlab testing - the tests which run on OL VMs (heavyvm), and the tests which run on vmcores.

heavyvm improvements:

- Drgn is now installed via an RPM rather than PyPI. This means that we're testing the exact OL drgn builds that customers would use, not the PyPI packages.
- Test outputs are now collected into an XML file, just like the vmcore tests.
- There are a few smaller improvements regarding how the test output looks, etc.

vmcore improvements:

- Vmcore tests can now be run in parallel!
- Uses the `CtfCompatibility` helper to check whether to run CTF tests.

Gitlab CI:

- We now run vmcore tests with CTF.
- Tests are now split into three groups: vmcore (CTF), vmcore (DWARF), and vm tests. These can each run in parallel, reducing runtime dramatically.
- Each of the vmcore tests also runs with parallelism also.

Runtime is down from ~23 minutes to ~7 minutes. No regressions found.


biger410 (approved):

brenns10:
Thank you!

PR #76, opened 2024-03-18
cpuinfo: Fix the numa-to-cpulist parsing in x86_get_cpu_info()

jianfenw:
In the cpuinfo module, x86_get_cpu_info() will print a list of cpus for each NUMA node. However, in the current implementation, all CPU cores are considered as NUMA node0's CPUs and printed accordingly. This piece of information is incorrect when checking the CPU information for a vmcore produced by a multi-node server.

This commit fixes it by providing a x86_get_per_node_cpulist() function to retrieve per-node cpulists and using it in x86_get_cpu_info(). Tested for uek4,5,6,7.

### DIFF
--- a/drgn_tools/cpuinfo.py
+++ b/drgn_tools/cpuinfo.py
@@ -6,6 +6,7 @@ Helper to view cpuinfo data
 import argparse
 from typing import Any
 from typing import Dict
+from typing import List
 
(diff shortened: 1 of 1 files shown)


brenns10 (approved):
Looks good, will merge pending internal CI!

PR #77, opened 2024-03-27
couple fixes for block and md helpers

biger410:
Couple fixes was added to make block and md helpers working better with upstream which will benefit future uek. Beside that one fix was adding "cpu" to inflight io output to help tell whether the cpu that request was issued from is in offline status.

### DIFF
--- a/drgn_tools/block.py
+++ b/drgn_tools/block.py
@@ -278,9 +278,9 @@ def rq_op_ef295ecf(rq: Object) -> str:
     # rq.cmf_flags: 8 bits for encoding the operation, and the remaining 24 for flags
     REQ_OP_BITS = 8
     op_mask = (1 << REQ_OP_BITS) - 1
-    req_opf = {
-        value: name for (name, value) in prog.type("enum req_opf").enumerators
-    }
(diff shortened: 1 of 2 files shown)


brenns10 (approved):
Thanks, all these fixes look well done. Since some of them are for UEK7, which we have in the test suite, yet no tests were failing, do you have any ideas of ways that we could test these helpers better?

biger410:
> Thanks, all these fixes look well done. Since some of them are for UEK7, which we have in the test suite, yet no tests were failing, do you have any ideas of ways that we could test these helpers better?

I think our CI test did great job. The 1st and 4th commits are for kernel v6.x, so only if CI test supports LUCI, that can be detected. The 2nd commit is for a special upstream kernel build which didn't have RQF_IO_STAT configured, this will not happen for uek. 3rd patch is a corner case which is only for io request in the plug list, it's other kernel bug which caused io pending in that list, it's very hard to get a vmcore with io in that list if no that bug. The 5th commit is adding a new "cpu" field to show cpu status which is triggered by a customer issue.

PR #78, opened 2024-04-01
Fix for the net helpers recently merged

brenns10:
From @pkannoju, there was a patch missing from the net pull request merged recently.


brenns10:
I ended up including these fixes in #79 because I needed to fix up the tests due to virtutil anyway.

PR #79, opened 2024-04-02
Fixes for recently merged helpers

brenns10:
The virtutil module contained some test failures:

- UEK4 does not contain the `cpuhp_state` variable, handle it gracefully
- UEK6 CTF seems to omit the type for one variable, causing a KeyError. This is definitely a CTF bug (and from my initial investigation, appears to be a bug in CTF _generation_ not the drgn CTF code). Work around it here to ensure tests pass.

The net module was missing some changes that had been applied as part of the code review, introduce them here.


biger410 (approved):
Looks good

brenns10:
Added one more fix for VmallocTotal with 5-level paging.

biger410 (approved):

PR #81, opened 2024-04-22
LSE-334: Add fallback method to locate locks in stack frame

imran-kn:
For cases, where frame["lock"] can't provide the lock address and gives
ObjectAbsent exception, this can be used as fallback method to obtain
lock addresses.
There can still be cases where for other kernels or other configurations
or for same kernel/configuration compiled with a different gcc versions,
the lock address may not be available at offsets used here, but the idea
here is that since we are mostly debugging released kernels built using
Jenkins pipelines, we should not see gcc or kernel configuration variation
and this helper will avoid need to manually check stack frames to locate
the locks when frame["lock"] method does not work.

### DIFF
--- a/drgn_tools/lock.py
+++ b/drgn_tools/lock.py
@@ -118,6 +118,8 @@ def scan_mutex_lock(
 ) -> None:
     """Scan for mutex and show details"""
     wtask = None
+    seen_mutex_waiters: Set[int] = set()
+
     if pid is not None:
         wtask = find_task(prog, pid)
 
@@ -129,23 +131,46 @@ def scan_mutex_lock(
 
     warned_absent = False
     for task, frame in frame_list:
+        if task.pid.value_() in seen_mutex_waiters:
+            continue
         try:
             mutex = frame["lock"]
-            mutex_addr = mutex.value_()
         except drgn.ObjectAbsentError:
-            if not warned_absent:
-                print(
-                    "warning: failed to get mutex from stack frame"
-                    "- information is incomplete"
-                )
-                warned_absent = True
+            mutex = get_lock_from_stack_frame(
+                prog, task.pid.value_(), frame, "mutex"
+            )
+            if not mutex:
+                task_frame_list = prog.stack_trace(task.pid.value_())
+                for tmp_frame in task_frame_list:
+                    if tmp_frame.name != "__schedule":
+                        continue
+
+                    mutex = get_lock_from_stack_frame(
+                        prog, task.pid.value_(), tmp_frame, "mutex"
+                    )
+                    if not mutex:
+                        if not warned_absent:
+                            print(
+                                "warning: failed to get mutex from stack frame"
+                                "- information is incomplete"
+                            )
+                            warned_absent = True
+
+        if not mutex:
             continue
 
         struct_owner = mutex_owner(prog, mutex)
 
+        mutex_addr = mutex.value_()
         if mutex_addr in seen_mutexes:
             continue
         seen_mutexes.add(mutex_addr)
+        seen_mutex_waiters.update(
+            [
+                waiter.pid.value_()
+                for waiter in for_each_mutex_waiter(prog, mutex)
+            ]
+        )
 
         index = 0
         print(f"Mutex: 0x{mutex_addr:x}")
--- a/drgn_tools/lock.py
+++ b/drgn_tools/lock.py
@@ -47,6 +47,7 @@ from drgn_tools.locking import _RWSEM_READER_MASK
 from drgn_tools.locking import _RWSEM_READER_SHIFT
 from drgn_tools.locking import for_each_mutex_waiter
 from drgn_tools.locking import for_each_rwsem_waiter
+from drgn_tools.locking import get_lock_from_stack_frame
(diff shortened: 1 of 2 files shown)


imran-kn:
The offsets used in these helpers were obtained using 2 kind of images:
1. UEK builds (UEK-5,6,7 and 8) from Jenkins.
2. Compiling UEK equivalent linux-stable kernels (v4.14, v5.4, v5.15 and v6.1) locally using gcc-8.5 to gcc-12.3. These locally compiled images were then used to boot and crash a qemu instance.

Also so far I have tested these only for x86_64 instances (local qemu instance  and OCI VM instances .

In worst case we will not be able to locate the lock, that we are chasing . But these helpers will never return a wrong lock because for the lock that has been located we obtain its waiter list and then check if the task whose stack frame we are investigating here is one of the waiters in that waiter list or not.
So even in worst case we will be at same situation , that we are at right now when frame["lock"] can't locate the lock

brenns10 (changes requested):
Sorry for a long delay in reviewing this. It's really exciting to see your work in this direction! I have a few reservations about merging this work as-is, so let me run them by you.

1. I may be missing the point of this, but I notice that many of the stack frames you're looking at are static inline functions (e.g. `__down_common` and `__mutex_lock`). These are functions which wouldn't appear in a stack trace unless you're using DWARF debuginfo. So these offsets are not really helpful for the case where you're using CTF, but still want to get lock information from a stack trace. My main concern with the lock commands is the lack of functionality when CTF debuginfo is enabled, but it seems like this PR doesn't address that. I guess you're looking to reduce number of cases where you get an `ObjectAbsentError` with DWARF debuginfo?
2. If this is only tested on `x86_64` then we really ought to avoid executing code like `frame.register('rbp')` on architectures which we know don't have that register (aarch64).
3. The `kmaj, jmin, _ = kernel_version(prog)` code is a bit of a kludge. I think your intention is to only support UEK kernels of specific versions, and I think the logic behind that is sound: this is very compiler/configuration specific. I think you'd be better off using the class `drgn_tools.debuginfo.KernelVersion` and using the `uek_version` field. That way you can explicitly say "this is only supported for UEK versions X, Y, Z". It also makes the code a bit simpler since you don't neeed "kmaj == 5 and kmin == 15".

Finally, my main concern is that while this code works now, there's a lot of context hidden in the _writing_ of the code. I think (a) it would be better to encode the stack offsets into dictionaries, rather than if statements. This makes the information _data_, not _code_. Then, (b) it would be really important to have documentation about how to generate that data! When UEK8 and UEK9 come out, we'll need to update those tables, and it would be best to document the process.

Ideally, we could build the stack offset tables using the DWARF debuginfo, e.g. using the `llvm-dwarfdump` to output the DWARF CFI information and determine at what offset the argument is pushed to the stack. (See bug 36564821 for details on that). That way we would be able to fully automate this ability.

imran-kn:
> Sorry for a long delay in reviewing this. It's really exciting to see your work in this direction! I have a few reservations about merging this work as-is, so let me run them by you.
> 
>     1. I may be missing the point of this, but I notice that many of the stack frames you're looking at are static inline functions (e.g. `__down_common` and `__mutex_lock`). These are functions which wouldn't appear in a stack trace unless you're using DWARF debuginfo. So these offsets are not really helpful for the case where you're using CTF, but still want to get lock information from a stack trace. My main concern with the lock commands is the lack of functionality when CTF debuginfo is enabled, but it seems like this PR doesn't address that. I guess you're looking to reduce number of cases where you get an `ObjectAbsentError` with DWARF debuginfo?
> 
>     2. If this is only tested on `x86_64` then we really ought to avoid executing code like `frame.register('rbp')` on architectures which we know don't have that register (aarch64).
> 
>     3. The `kmaj, jmin, _ = kernel_version(prog)` code is a bit of a kludge. I think your intention is to only support UEK kernels of specific versions, and I think the logic behind that is sound: this is very compiler/configuration specific. I think you'd be better off using the class `drgn_tools.debuginfo.KernelVersion` and using the `uek_version` field. That way you can explicitly say "this is only supported for UEK versions X, Y, Z". It also makes the code a bit simpler since you don't neeed "kmaj == 5 and kmin == 15".
> 
> 
> Finally, my main concern is that while this code works now, there's a lot of context hidden in the _writing_ of the code. I think (a) it would be better to encode the stack offsets into dictionaries, rather than if statements. This makes the information _data_, not _code_. Then, (b) it would be really important to have documentation about how to generate that data! When UEK8 and UEK9 come out, we'll need to update those tables, and it would be best to document the process.
> 
> Ideally, we could build the stack offset tables using the DWARF debuginfo, e.g. using the `llvm-dwarfdump` to output the DWARF CFI information and determine at what offset the argument is pushed to the stack. (See bug 36564821 for details on that). That way we would be able to fully automate this ability.

Thanks @brenns10 for taking a look. I had my apprehensions with this PR but I also wanted to have a starting which we have now. some of these concerns like using frame.sp everywhere and avoid kmaj/kmin have already been addressed in my local copy . I will address other concerns and get back on this.

brenns10:
Closing this, as it was superseded by #99. Thanks for the work on this!

PR #85, opened 2024-05-30
Github actions fix, UEK-next support, and a table fix

brenns10:
There are three separate changes here, sorry. But each is small and easy to understand.

1. Add support for UEK-next, so corelens will run against it. Also add uek-next to the "lite VM" tests, but don't enable it yet because some tests fail. I've cataloged all the failures.
2. Get rid of caching in github actions, it doesn't save any time and it causes more failures.
3. Fix alignment of some tables.


biger410 (approved):
Looks good.

PR #86, opened 2024-05-31
LSE-367: irq: fix for v6.5 and later kernels.

imran-kn:
v6.5 moved irq descriptors into a maple tree,
earlier they used to exist in a radix tree.
Change irq helpers to accomodate this change.

LSE-367.


brenns10 (approved):
All IRQ tests pass on UEK-next with this fix. Code looks good, thank you!

PR #87, opened 2024-05-31
LSE-373: workqueue: Fix errors seen for newer (v6.4 and later) kernels.

imran-kn:
LSE-373.

### DIFF
--- a/drgn_tools/workqueue.py
+++ b/drgn_tools/workqueue.py
@@ -12,6 +12,7 @@ from typing import Optional
 from typing import Union
 
 from drgn import cast
+from drgn import IntegerLike
 from drgn import NULL
 from drgn import Object
 from drgn import Program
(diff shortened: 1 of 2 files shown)


drgn_tools/workqueue.py
```diff
@@ -63,6 +94,33 @@ def _print_work(work: Object) -> None:
...
+    # we can track both name change and type change of this member.
+    try:
+        pwq = per_cpu_ptr(workqueue.cpu_pwqs, cpu)
+    except AttributeError:
+        prog = workqueue.prog_
+        pwq = Object(
+            prog,
+            "struct pool_workqueue",
+            address=prog.read_word(per_cpu_ptr(workqueue.cpu_pwq, cpu)),
+        ).address_of_()
```
brenns10, line 119: You can simplify this by just dereferencing the double pointer.
```suggestion
        pwq = per_cpu_ptr(workqueue.cpu_pwq, cpu)[0]
```

brenns10 (changes requested):
I tested this on UEK-next and it resolves all the workqueue related issues. Looks to pass on gitlab CI too. However, one tiny nit if you don't mind. Thanks!

drgn_tools/workqueue.py
imran-kn: Makes sense. Have done that now.

brenns10 (approved):
Thanks for the change, everything else is good! The test failure is unrelated to your change.

PR #88, opened 2024-05-31
LSE-368: tests: Fix test failures due to movement of cpu_number.

imran-kn:
cpu_number has been merged into pcpu_hot object, for x86, in newer (v6.2+) kernels. For some archs like aarch64, it still exists as per-cpu variable.

Since we just need a per-cpu address for this test, use per-cpu runqueues. These have been existing as per-cpu varibale for long time and even if this changes in future the change will be architecture independent, so we would not have to worry about test passing for one arch and failing for other.

LSE-368.

### DIFF
--- a/tests/test_mm.py
+++ b/tests/test_mm.py
@@ -63,7 +63,7 @@ def test_AddrKind_categorize_data(prog: drgn.Program) -> None:
     assert mm.AddrKind.categorize(prog, bss) == mm.AddrKind.BSS
 
     # percpu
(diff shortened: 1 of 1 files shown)


brenns10 (approved):

brenns10:
Thanks, I wrote this test and should have recognized the issue immediately! I guess I was too caught up in filing tickets. Anyway, I tested this on UEK-next and it eliminated the reported errors. Thank you!

PR #89, opened 2024-06-07
rds_v2: Added str queues and rds_ib_mr info for busy/clean lists

RN97:
Added new functions that display the rds send, retransmit and receive queues. Also added functions that display the MR list info for busy and clean lists.

### DIFF
--- a/drgn_tools/rds.py
+++ b/drgn_tools/rds.py
@@ -22,10 +22,13 @@ import drgn
 from drgn import cast
 from drgn import container_of
 from drgn import Object
+from drgn import PlatformFlags
 from drgn import Program
 from drgn.helpers.linux import for_each_online_cpu
 from drgn.helpers.linux import per_cpu
 from drgn.helpers.linux.list import hlist_for_each_entry
+from drgn.helpers.linux.list import list_empty
+from drgn.helpers.linux.list import list_for_each
 from drgn.helpers.linux.list import list_for_each_entry
 from drgn.helpers.linux.pid import find_task
 
@@ -80,6 +83,51 @@ RDMA_CM_STATES = {
 }
 
 # Helpers #
+
+
+def be64_to_host(prog: drgn.Program, value: int) -> int:
+    """
+    Convert 64 byte value from big endian to host order
+
+    :param prog: drgn program
+    :param value: The value to be converted
+    """
+    # If the platform byte order differs from ours, drgn will
+    # transparently handle that. We only need to do a byte
+    # swap if the program byte order is little endian.
+    if prog.platform.flags & PlatformFlags.IS_LITTLE_ENDIAN:
+        return struct.unpack("<Q", struct.pack(">Q", value))[0]
+    return value
+
+
+def be32_to_host(prog: drgn.Program, value: int) -> int:
+    """
+    Convert 32 byte value from big endian to host order
+
+    :param prog: drgn program
+    :param value: The value to be converted
+    """
+    # If the platform byte order differs from ours, drgn will
+    # transparently handle that. We only need to do a byte
+    # swap if the program byte order is little endian.
+    if prog.platform.flags & PlatformFlags.IS_LITTLE_ENDIAN:
+        return struct.unpack("<L", struct.pack(">L", value))[0]
+    return value
+
+
+def be16_to_host(prog: drgn.Program, value: int) -> int:
+    """
+    Convert 16 byte value from big endian to host order
+
+    :param prog: drgn program
+    :param value: The value to be converted
+    """
+    # If the platform byte order differs from ours, drgn will
+    # transparently handle that. We only need to do a byte
+    # swap if the program byte order is little endian.
+    if prog.platform.flags & PlatformFlags.IS_LITTLE_ENDIAN:
+        return struct.unpack("<H", struct.pack(">H", value))[0]
+    return value
 
 
 def rds_inet_ntoa(addr_obj: Object) -> str:
@@ -429,6 +477,33 @@ def rds_ip_state_list(dev: Object) -> Dict[str, Tuple[int, int]]:
         else:
             ip_list[ipaddr] = (num_conns_up, num_conns + 1)
     return ip_list
+
+
+def ib_mr_type(ib_mr: Object) -> str:
+    """
+    Get the IB_MR_TYPE for a given RDS IB_MR.
+
+    :param ib_mr: ``struct rdma_id_private`` Object.
+    :returns: The RDS connection RDMA CM state in string format.
+    """
+
+    ib_mr_types = {
+        0: "IB_MR_TYPE_MEM_REG",
+        1: "IB_MR_TYPE_SG_GAPS",
+        2: "IB_MR_TYPE_DM",
+        3: "IB_MR_TYPE_USER",
+        4: "IB_MR_TYPE_DMA",
+        5: "IB_MR_TYPE_INTEGRITY",
+    }
+
+    if ib_mr is None:
+        return "N/A"
+
+    key = ib_mr.type.value_()
+    if key not in ib_mr_types.keys():
+        return "N/A"
+
+    return ib_mr_types[key]
 
 
 # RDS Corelens module functions #
@@ -786,6 +861,8 @@ def rds_info_verbose(
             "Rx_poll_cnt",
             "SCQ vector",
             "RCQ vector",
+            "SND IRQN",
+            "RCV IRQN",
         ]
     ]
     for dev in for_each_rds_ib_device(prog):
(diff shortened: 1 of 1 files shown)


drgn_tools/rds.py
```diff
@@ -1020,6 +1117,421 @@ def rds_sock_info(
...
+            " ",
+            "LocalAddr",
+            "LPort",
+            "RemoteAddr",
+            "RPort",
+            "Tos",
+            "Seq",
+            "Bytes",
+        ]
+    ]
```
brenns10, line 1161: It would be cool to start using `Table` rather than `print_table()`. It's very similar but more flexible. And if necessary, it can be converted to have immediate output (at the expense of possible misaligned columns).

```suggestion
    print("Receive message queue:")
    msg_queue = Table(
        [
            " ",
            "LocalAddr",
            "LPort",
            "RemoteAddr",
            "RPort",
            "Tos",
            "Seq",
            "Bytes",
        ],
        outfile=outfile,
        report=report,
    )
```

drgn_tools/rds.py
```diff
@@ -1020,6 +1117,421 @@ def rds_sock_info(
...
+                    index,
+                    local_addr,
+                    lport_rm,
+                    remote_addr,
+                    rport_rm,
+                    tos_rm,
+                    seq,
+                    num_bytes,
+                ]
+            )
```
brenns10, line 1213:
```suggestion
            msg_queue.row(
                index,
                local_addr,
                lport_rm,
                remote_addr,
                rport_rm,
                tos_rm,
                seq,
                num_bytes,
            )
```

drgn_tools/rds.py
```diff
@@ -1020,6 +1117,421 @@ def rds_sock_info(
...
+
+    print("Receive Message Queue:")
+    print_table(msg_queue, outfile, report)
+    print("\n")
```
brenns10, line 1217:
```suggestion
    msg_queue.write()
```

drgn_tools/rds.py
```diff
@@ -1020,6 +1117,421 @@ def rds_sock_info(
...
+    :param report: Open the file in append mode. Used to generate a report of all the functions in the rds module.
+    :returns: None
+
+    """
+    msg = ensure_debuginfo(prog, ["rds"])
+    if msg:
+        print(msg)
+
+    index = -1
+    msg_queue: List[List[Any]] = [
```
brenns10, line 1253: Would be nice to use `Table()` here as well, following the same general approac above.

drgn_tools/rds.py
```diff
@@ -1020,6 +1117,421 @@ def rds_sock_info(
...
+    ret: Optional[bool] = False,
+    outfile: Optional[str] = None,
+    report: bool = False,
+) -> None:
+    """
+    Print the rds send or retransmit msg queue similar rds-info -st
+
+    :param prog: drgn program
+    :prarm queue: The msg queue to be displaed. Ex: 'send', 'recv'
+    :param laddr: comma seperated string list of LOCAL-IP.  Ex: '192.168.X.X, 10.211.X.X, ...'
```
brenns10, line 1238: spelling nit: separated

drgn_tools/rds.py
```diff
@@ -1020,6 +1117,421 @@ def rds_sock_info(
...
+    :param lport: comma seperated string list of lport.  Ex: 2259, 36554, ...'
+    :param rport: comma seperated string list of rport.  Ex: 2259, 36554, ...'
+    :param outfile: A file to write the output to.
+    :param report: Open the file in append mode. Used to generate a report of all the functions in the rds module.
+    :returns: None
+
+    """
+    msg = ensure_debuginfo(prog, ["rds"])
+    if msg:
+        print(msg)
```
brenns10, line 1372: Do not use `ensure_debuginfo()` inside a corelens module code

drgn_tools/rds.py
```diff
@@ -1020,6 +1117,421 @@ def rds_sock_info(
...
+            ret,
+            outfile,
+            report,
+        )
+    elif queue.lower() == "recv" or queue.lower() == "rcv":
+        rds_print_recv_msg_queue(
+            prog, laddr, raddr, tos, lport, rport, ret, outfile, report
+        )
+    else:
+        print("Incorrect queue arg provided !")
```
brenns10, line 1415: Let's reduce this a bit:
```suggestion
    queue = queue.lower()
    if queue not in ("all", "send", "snd", "retrans", "re", "recv", "rcv"):
        print(f"Unknown queue type '{queue}'. Expected: all, send, retrans, or recv")
        return

    if queue in ("send", "snd", "all"):
        rds_print_send_retrans_msg_queue(
            prog, "send", laddr, raddr, tos, lport, rport, ret, outfile, report
        )
    if queue in ("retrans", "re", "all"):
        rds_print_send_retrans_msg_queue(
            prog,
            "retrans",
            laddr,
            raddr,
            tos,
            lport,
            rport,
            ret,
            outfile,
            report,
        )
    if queue in ("recv", "rcv", "all"):
        rds_print_recv_msg_queue(
            prog, laddr, raddr, tos, lport, rport, ret, outfile, report
        )
```

drgn_tools/rds.py
```diff
@@ -1020,6 +1117,421 @@ def rds_sock_info(
...
+    :param list_name: The name of the list. Ex 'busy_list', 'clean_list'
+    :returns: None
+
+    """
+    if list_empty(list_head.address_of_()):
+        print("\nrds_ib_mr->{}->{} is empty!\n".format(pool_name, list_name))
+        return
+    index = 0
+
+    list_info: List[List[Any]] = [
```
brenns10, line 1438: Another `Table` opportunity :)

drgn_tools/rds.py
```diff
@@ -1039,6 +1551,7 @@ def report(prog: drgn.Program, outfile: Optional[str] = None) -> None:
     rds_conn_info(prog, outfile=outfile, report=True)
     rds_info_verbose(prog, outfile=outfile, report=True)
     rds_stats(prog, outfile=outfile, report=True)
+    rds_print_msg_queue(prog, queue="All", outfile=outfile, report=True)
 
 
 class Rds(CorelensModule):
```
brenns10, line 1557: So, you've started using structures from the `mlx_core` module.

Since `mlx_core` seems to be not _required_ you should add that module to the `debuginfo_kmods` list for the RDS corelens module. This won't require that module to be present, but it will require the module to be loaded if it is present.

Then you should also handle the case where the mlx-related structure debuginfo is not available.

brenns10 (changes requested):

drgn_tools/rds.py
RN97: Hey Stephen, is there any specific reason for this ? 

- When authoring the code, we had discussed that it was best practice to ensure the rds module in the functions. Has this changed ?

brenns10: I have two minor issues with ensure_debuginfo():

1. It's not 100% accurate at telling whether a module has debuginfo loaded. That can't really be helped though, and in practice it works well enough.
2. Every time the function is called, we have to iterate over the list of kernel modules in the vmcore. Even if debuginfo is loaded for that module. This may not be cheap, especially on live systems. So the first priority with ensure_debuginfo() is to never call it in a loop, or in small helper functions that get called repeatedly.

Overall, it's not terrible but I worry about unnecessary performance issues. But that's not really your problem in this code, so I'll drop the issue. I think it would be good to cache the result of ensure_debuginfo() to make it less of an issue when calls are repeated.

drgn_tools/rds.py
```diff
@@ -431,6 +480,56 @@ def rds_ip_state_list(dev: Object) -> Dict[str, Tuple[int, int]]:
...
+    else:
+        print("The device provided is not a MLX device.")
+        return False
+
+    msg = ensure_debuginfo(prog, module)
+    if msg:
+        print(msg)
+        return False
+
+    return True
```
brenns10, line 530: 1. So, like I shared in the previous review, the issue with `ensure_debuginfo()` is that each time it is called, it iterates over all kernel modules. I'll work on changing this, but it would be really nice to avoid doing two calls when you could do one. This is split out into two functions, but they are only ever called together. Could you make them a single function that does `ensure_debuginfo(prog, ["mlx5_core", "mlx5_ib"])` or the mlx4 variant? That would reduce the overhead.
2. The `print()` statements here are going to produce undesirable output in your tables. I'd argue against using print statements at all, you can just put "n/a" in the table rows as you're doing now.

drgn_tools/rds.py
```diff
@@ -875,6 +977,41 @@ def rds_info_verbose(
             except AttributeError:
                 rcq_vector = "N/A"
 
+            if ensure_mlx_core_debuginfo(
+                prog, dev_name
+            ) and ensure_mlx_ib_debuginfo(prog, dev_name):
```
brenns10, line 982: The result of these functions does not change at all throughout this inner loop (for each connection on the device) yet we're calling these ensure debuginfo functions every single iteration.

Let's move them out of the inner loop, at the very least. There can be a lot of connections on a single device and that will multiple to many unnecessary calls.

drgn_tools/rds.py
```diff
@@ -1020,6 +1159,390 @@ def rds_sock_info(
...
+            outfile,
+            report,
+        )
+        print("\n")
+    if queue in ("recv", "rcv", "all"):
+        rds_print_recv_msg_queue(
+            prog, laddr, raddr, tos, lport, rport, ret, outfile, report
+        )
+
+    return
```
brenns10, line 1432: nit: return here is unnecessary

brenns10 (changes requested):
Just a few smaller concerns.

brenns10 (approved):
Looks good, will merge after the tests pass.

PR #91, opened 2024-06-28
Couple dm and IO enhancements

biger410:
The following are included in the patch set:

1. For dm table, adding multipath and snapshort support
2. New fields adding to blockinfo helpers to troubleshooting io hung with queue freezed
3. couple helpers to remove redundant code or fix small issues

### DIFF
--- a/drgn_tools/dm.py
+++ b/drgn_tools/dm.py
@@ -10,6 +10,7 @@ from typing import Tuple
 from drgn import cast
 from drgn import Object
 from drgn import Program
+from drgn.helpers.linux.list import list_first_entry_or_null
 from drgn.helpers.linux.list import list_for_each_entry
 from drgn.helpers.linux.rbtree import rbtree_inorder_for_each_entry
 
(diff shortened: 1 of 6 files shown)


drgn_tools/dm.py
```diff
@@ -14,13 +14,37 @@
...
 
+def dm_ro(dm: Object) -> bool:
+    if has_member(dm, "bdev"):
+        bdev = dm.bdev
+    else:
+        bdev = dm.disk.part0
+    ro = blkdev_ro(bdev)
+    if ro == -1:
+        gendisk = dm.disk
+        return gendisk.part0.policy
```
brenns10, line 37: nit: this isn't a bool, it's a drgn object. Can you return `bool(gendisk.part0.policy)`?

drgn_tools/dm.py
```diff
@@ -258,9 +258,38 @@ def show_table_multipath(dm: Object, name: str) -> None:
...
+    if msg:
+        print(msg)
+        return
+
+    table = dm_table(dm)
+    for tid in range(table.num_targets):
+        target = table.targets[tid]
+        snapshot = cast("struct dm_snapshot *", target.private)
+        print(
+            "%s: %d %d %s %s %s %s %s %s %d"
```
brenns10, line 272: I'm not terribly familiar, does this need a table header? If so, it would be good to use the helpers from `drgn_tools/table.py` to print a header. As it is, I'm not sure if I'd know what the fields represent. But maybe this is similar to an existing format the kernel already outputs? Up to you whether you think this output is suitable.

brenns10 (changes requested):
Reviewed each commit individually. There's two tiny requests, nothing major. Thank you!

drgn_tools/dm.py
biger410: Unfortunately there isn't a fixed table header for this, different target driver type have different output, the first 4 columns will be in the same format, the following will be dynamic depending on the target driver type in the 4th column. And yes this is to simulate the the cmd output of `dmsetup table` .

brenns10: Ok, sounds good then.

brenns10 (approved):

PR #97, opened 2024-08-06
nfs_tools: fix problem of displaying string representation of tk_runsstate

daimngo:
This fix ensures the correct string representation of the bits in tk_runstate, which vary depending on the kernel version.

### DIFF
--- a/drgn_tools/nfs_tools.py
+++ b/drgn_tools/nfs_tools.py
@@ -40,12 +40,79 @@ __all__ = (
 """ include/linux/sunrpc/sched.h """
 
 
-class RpcTaskState(BitNumberFlags):
+class RpcTaskState_1(BitNumberFlags):
+    RPC_TASK_RUNNING = 0
+    RPC_TASK_QUEUED = 1
+    RPC_TASK_ACTIVE = 2
+
+
+# 729749bb8da1 SUNRPC: Don't hold the transport lock across socket copy operations
+class RpcTaskState_2(BitNumberFlags):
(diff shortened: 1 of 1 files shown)


drgn_tools/nfs_tools.py
```diff
@@ -40,14 +40,81 @@
...
+    RPC_TASK_MSG_PIN_WAIT = 5
+    RPC_TASK_SIGNALLED = 6
+
+
+def does_func_exist(prog: drgn.Program, func: str) -> bool:
+    try:
+        _ = prog[func]
+        return True
+    except KeyError:
+        return False
```
brenns10, line 95: This doesn't actually work for CTF, since CTF doesn't always contain the type definitions of functions. So looking up a function as a _drgn object_ doesn't reliably work with CTF. However, you can look up the corresponding _symbol_ which will be there for both CTF and DWARF.
```suggestion
    try:
        prog.symbol(func)
        return True
    except LookupError:
        return False
```

brenns10 (changes requested):
One small change requested here due to weirdness with debuginfo. But otherwise this looks excellent, thanks for finding and fixing this.

drgn_tools/nfs_tools.py
daimngo: Thanks Stephen, I made the suggested changes.

brenns10: I've retried it, I'm sure it was just a transient error, no worries!

brenns10 (approved):

PR #99, opened 2024-08-12
LSE-334: Fallback for getting locks from stack frames

brenns10:
This pull request augments the existing lock module. Previously, it could only work with DWARF debuginfo, using the CFI to identify the mutex, sem, or rwsem that a task was blocked waiting on. Now, it can also function in cases where the DWARF debuginfo fails, as a fallback. What's more, it also works with CTF debuginfo.

The approach here is based on @imran-kn's idea that you can guess that a pointer corresponds to a lock, and then you can verify that by checking to see whether the current task is on the list of waiters for that lock. The catch here is that you need to be really careful about it: it's not enough to just do a "try/except" for FaultErrors. You also need to ensure that the linked list nodes are valid (so it's less likely that you're dereferencing random pointers). And even if the linked list nodes are valid, you could end up with an infinite loop, so we also need to track the set of already visited nodes in the list. If we do all this, we can be confident that either we will (a) dereference some junk and generate a FaultError, (b) encounter a pointer to something that's not a valid linked list node, so we'll get a ValidationError, or (c) we'll make a cycle in the list and raise a ValidationError because that's not valid data either.

With the _careful_ checking described above, there's no need to have complex tables of offsets! We just need to find the start and stop location on the stack, and search every word on the stack for a valid lock. We can speed things along by checking what kind of address we're looking at, and only testing the pointer if it could be a valid lock pointer (skipping userspace addresses, text addresses, etc). This was @biger410's idea, if he hadn't mentioned it, I would still be generating tables of offsets...

To test this, I've created a large crop of vmcores with kernel modules loaded. These modules spawn several kthreads that wait for a lock, and then the test case can find these threads and assert that it is able to find the lock from the stack frame, both with and without the fallback.

---

Tests pass on all vmcores internally (including the newly uploaded ones). Tests pass on Github CI as well, though there's not much testing of this specific functionality.

I think this will probably replace #81.


drgn_tools/locking.py
```diff
@@ -519,3 +570,92 @@ def get_rwsem_info(rwsem: Object, callstack: int = 0) -> None:
...
+
+    # Fall back to a brute force method which works shockingly well. Simply scan
+    # from the top of the stack down to the frame of the lock. If the address is
+    # a kernel address, we'll test it (carefully) to see whether it's a valid
+    # lock, and if so, we'll return it.
+    #
+    # First, we need to get the start and end address to search. This is
+    # arch-specific.
+    if prog.platform.arch == Architecture.X86_64:
+        candidates = range(task.thread.sp, frame.sp, 8)
```
biger410, line 633: `frame.sp` is the bottom of the stack?

brenns10: Yeah, it's the bottom of this stack frame. This reliably works well though! I think it's because x86 always spills the register beneath the stack frame anyway. With aarch64, it seems that the register gets stored within the stack frame.

I could try to get the next stack frame up, but this function doesn't actually accept a stack trace... just the single stack frame. I could try to use the workaround with the frame pointer that aarch64 does. In practice, we always find the lock before we get to the upper end up the range anyway, so it wouldn't be much of an issue.

biger410: so `task.thread.sp` is the bottom of the stack, then the scan happened to the parents stack frame of current one, could there be a case that lock address is stored in the children stack frame of current one?

brenns10: I guess I'm not sure what you mean. The `task.thread.sp` is the "bottom" of the stack, that is, it is where the stack pointer was stored just prior to context switching away. Then, `frame.sp` is the stack pointer for the current function (e.g. `mutex_lock()`). So, the stack looks something like this:

```
HIGHER ADDRESSES / OLDER FRAMES HERE
[parent caller frame]
...
[mutex_lock() frame]
0xDEADBEEF
0xF000B000
... vars ...                                       <---- frame.sp
[schedule_preempt_disabled() frame]
...
[schedule() frame]
... vars ...
... bottom of the stack ...                     <--- task.thread.sp
LOWER ADDRESSES / NEWER FRAMES HERE
```

So we start at `task.thread.sp` and increment the address until we reach `frame.sp`. Since `frame.sp` is technically pointing at the bottom of the stack frame for `mutex_lock()`, it's possible that the lock could be somewhere in that frame and we'd miss it. However in practice that just doesn't happen, the lock has always been beneath the frame.sp value.

biger410: My fault, thought `task.thread.sp` is the other end of the stack. 
I'm OK with current approach, we can always fix it if lock is stored in current stack frame in future kernels.

biger410 (approved):

brenns10:
Thanks @biger410! I'll leave this open until Monday morning PST so @imran-kn can have a chance to raise any issues (since some of the code is his), but I'll merge if I don't hear anything. It would be great to get the 1.0 release branch started!

imran-kn (approved):
LGTM. Super cool stuff @brenns10  : )

PR #103, opened 2024-08-27
A slabinfo fix, and the v1 (1.1.0) release

brenns10:
The slabinfo fix was more involved than I wanted, but it works well, is compatible with drgn 0.0.25 - 0.0.27, and it actually makes the `slabinfo` module modestly faster.

### DIFF
--- a/buildrpm/python-drgn-tools.spec
+++ b/buildrpm/python-drgn-tools.spec
@@ -1,7 +1,7 @@
 # Copyright (c) 2024, Oracle and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
 Name:           python-drgn-tools
-Version:        0.9.1
+Version:        1.1.0
 Release:        1%{?dist}
 Summary:        Helper scripts for drgn, containing the corelens utility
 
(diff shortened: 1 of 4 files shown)


buildrpm/python-drgn-tools.spec
```diff
@@ -1,7 +1,7 @@
 # Copyright (c) 2024, Oracle and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
 Name:           python-drgn-tools
-Version:        0.9.1
+Version:        1.0.0
```
biger410, line 4: Here is the main branch, so it is still 1.0.0? You will create a new branch for v1 and set version to 1.1.0 there and also change the version here to 2.0.0, right?

brenns10: No, I totally missed that. This should go to 1.1.0, I'll create the `stable/v1` branch, and then on the main branch I'll bump to 2.0.0.

buildrpm/python-drgn-tools.spec
```diff
@@ -61,6 +61,9 @@ rm %{buildroot}/usr/bin/DRGN
 %{_mandir}/man1/corelens.1.gz
 
 %changelog
+* Tue Aug 27 2024 Stephen Brennan <stephen.s.brennan@oracle.com> - 1.1.0-1
```
biger410, line 64: should this be 1.1.0 instead of 1.1.0-1?

brenns10: No, most RPMs have a release version appended with the hyphen, it starts at 1. If we update the packaging alone, we'll update that number.

brenns10: It comes from line 5 above.

biger410 (approved):
Looks good.

PR #104, opened 2024-09-11
runq: add prio and runtime

richl9:
Dump prio and runtime as well. Removed  min_run_time_seconds parameter since we have another lockup helper pending to merge.

### DIFF
--- a/drgn_tools/runq.py
+++ b/drgn_tools/runq.py
@@ -67,13 +67,12 @@ def _print_cfs_runq(runqueue: Object) -> None:
         print("     [no tasks queued]")
 
 
-def run_queue(prog: Program, min_run_time_seconds: int = 0) -> None:
+def run_queue(prog: Program) -> None:
     """
-    Print tasks which are in the RT and CFS runqueues on each CPU. Specify min_run_time_seconds to x to only print
+    Print tasks which are in the RT and CFS runqueues on each CPU.
     processes running more than x seconds.
 
     :param prog: drgn program
-    :param min_run_time_seconds: int
     """
 
     # _cpu = drgn.helpers.linux.cpumask.for_each_online_cpu(prog)
@@ -85,8 +84,6 @@ def run_queue(prog: Program, min_run_time_seconds: int = 0) -> None:
         pid = curr_task.pid.value_()
         run_time = task_lastrun2now(curr_task)
         prio = curr_task.prio.value_()
-        if run_time < min_run_time_seconds * 1e9:
-            continue
         print(f"CPU {cpus} RUNQUEUE: {runqueue.address_of_().value_():x}")
         print(
             f"  CURRENT:   PID: {pid:<6d}  TASK: {curr_task_addr:x}  PRIO: {prio}"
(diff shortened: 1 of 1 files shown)


drgn_tools/runq.py
```diff
@@ -65,20 +67,31 @@ def _print_cfs_runq(runqueue: Object) -> None:
...
+    :param min_run_time_seconds: int
     """
 
     # _cpu = drgn.helpers.linux.cpumask.for_each_online_cpu(prog)
     for cpus in for_each_online_cpu(prog):
         runqueue = per_cpu(prog["runqueues"], cpus)
-        comm = escape_ascii_string(runqueue.curr.comm.string_())
-        pid = runqueue.curr.pid.value_()
+        curr_task_addr = runqueue.curr.value_()
+        curr_task = runqueue.curr[0]
```
biger410, line 83: Why use `runqueue.curr[0]`, `runqueue.curr` not working?

drgn_tools/runq.py
```diff
@@ -65,20 +67,31 @@ def _print_cfs_runq(runqueue: Object) -> None:
...
+    :param prog: drgn program
+    :param min_run_time_seconds: int
     """
 
     # _cpu = drgn.helpers.linux.cpumask.for_each_online_cpu(prog)
     for cpus in for_each_online_cpu(prog):
         runqueue = per_cpu(prog["runqueues"], cpus)
-        comm = escape_ascii_string(runqueue.curr.comm.string_())
-        pid = runqueue.curr.pid.value_()
+        curr_task_addr = runqueue.curr.value_()
```
biger410, line 82: This variable is only use when print, can we remove it and just use `curr_task.value_()`?

drgn_tools/runq.py
richl9: This change is due to a bug we observed . The pointer object 'runqueue.curr' can change in between when we are getting its attributes, so we need to dereference it first to get the value object for safety.

drgn_tools/runq.py
richl9: For the same reason, since 'runqueue.curr' could change, we want to get its current address first, followed by de-referencing it via runqueue.curr[0] immediately.

drgn_tools/runq.py
biger410: i see

drgn_tools/runq.py
```diff
@@ -65,20 +67,31 @@ def _print_cfs_runq(runqueue: Object) -> None:
         print("     [no tasks queued]")
 
 
-def run_queue(prog: Program) -> None:
+def run_queue(prog: Program, min_run_time_seconds: int = 0) -> None:
```
biger410, line 70: I see you have another lockup helper to down current task in the runqueue which has been run for long time, and it didn't use this helper, then why we need the second parameter `min_run_time_seconds` here?

richl9: You are right. Let me remove the parameter since now we have a  lockup helper.

biger410 (approved):
Looks good to me.

PR #105, opened 2024-09-18
fixed dm and Makefile

richl9:


Makefile
```diff
@@ -33,4 +33,4 @@ drgn_tools/_version.py:
 .PHONY: rsync
 rsync: drgn_tools/_version.py
 	@if [ -z "$(TARGET)" ]; then echo "error: TARGET unspecified. Either set it in config.mk, or use\nmake TARGET=hostname rsync"; exit 1; fi
-	rsync -avz --exclude "__pycache__" --exclude ".git" --exclude ".mypy_cache" ./drgn_tools $(TARGET):drgn_tools/
+	rsync -avz --exclude "__pycache__" --exclude ".git" --exclude ".mypy_cache" ./drgn_tools $(TARGET)
```
brenns10, line 36: To ensure that this doesn't break any existing users who may already have TARGET set and expect it to go to `TARGET:drgn_tools/`, would you please go ahead and add the following code into the Makefile? Directly beneath the `-include config.mk` directive would be best.
```makefile
# The TARGET variable could be either a hostname or a hostname:path. Previously
# the makefile expected just a hostname, and it always placed the resulting
# directory at $TARGET:drgn_tools/. But it's nice to have flexibility to change
# the target directory, so we only add that default path if there's not already
# a path specified.
ifneq (,$(TARGET))
ifeq (,$(findstring :,$(TARGET)))
override TARGET := $(TARGET):drgn_tools/
else
endif
endif
```

drgn_tools/dm.py
```diff
@@ -26,14 +26,22 @@ def for_each_dm_hash(prog: Program) -> Iterable[Tuple[Object, str, str]]:
...
 def for_each_dm_rbtree(prog: Program) -> Iterable[Tuple[Object, str, str]]:
     for hc in rbtree_inorder_for_each_entry(
         "struct hash_cell", prog["name_rb_tree"], "name_node"
     ):
-        yield hc.md, hc.name.string_().decode(), hc.uuid.string_().decode()
+        uuid = ""
+        if hc.uuid:
+            uuid = hc.uuid.string_().decode()
+
+        yield hc.md, hc.name.string_().decode(), uuid
```
brenns10, line 44: The fixes look good, but the commit message doesn't contain any context for why they are necessary. Was there a change that made the `uuid` field nullable, or has it always been and we just didn't handle it?

brenns10 (changes requested):
A few things in addition to the two comments:

1. Can you file an (Oracle) bug for this change? And can you say whether the dm fix should be backported to v1? If so, let's file both bugs for that.
2. I see the merge commit in there, I try to avoid merge commits wherever possible. Could you get rid of that?
3. Would you mind splitting out the `dm.py` change and the `Makefile` commits into their own separate commits? I don't mind them being in the same pull request, but they are logically different changes and shouldn't live in the same commit.

PR #107, opened 2024-09-19
corelens: sysctl: prints out sysctl entries

richl9:
add sysctl to corelens and cover more corner cases

### DIFF
--- a/drgn_tools/sysctl.py
+++ b/drgn_tools/sysctl.py
@@ -1,8 +1,9 @@
 # Copyright (c) 2023, Oracle and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
 """
-Helpers for sysctl.
+Helpers for sysctl. Further improvements are still needed to cover more corner cases and stay consistent with sysctl command output.
 """
+import argparse
 from stat import S_ISDIR
 from stat import S_ISLNK
 from types import MappingProxyType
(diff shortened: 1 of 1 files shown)


drgn_tools/sysctl.py
```diff
@@ -18,11 +19,19 @@
...
 
+from drgn_tools.corelens import CorelensModule
 
-def get_sysctl_table(prog: Program) -> MappingProxyType:
+
+MAX_ELEMENTS = 100
+
+
+def get_sysctl_table(
+    prog: Program, limit: int = MAX_ELEMENTS
```
brenns10, line 29: Adding a `limit` parameter here breaks the caching. Suppose this function gets called with `limit=50`, and that result gets cached, and then later it is called with `limit=100`. The cached result is returned and it only has 50 entries.

Is there a use case for this yet? E.g. a particular sysctl that needs this? If not, I'd recommend removing the logic. But if there's a use case, let's take the `limit` and apply it in `print_sysctl_table()` rather than in `get_data_entry()`. So `get_data_entry()` would get the full value, and then the `print_sysctl_table()` would decide how much of it to print.

brenns10 (changes requested):
Thanks for adding the corelens module for this. It looks like the PR also contains your DM changes and the makefile changes, so it needs an update (rebase).

Also be sure to file a bug for this, I think it makes sense to keep this in `main` only, with no stable backport.

drgn_tools/sysctl.py
richl9: Right. I don't think it's of much use. Removed it and made it just dump all.

brenns10 (approved):

PR #108, opened 2024-09-30
Add iscsi helpers to dump iscsi session information

richl9:
Updated the helper based on feedback from @biger410

### DIFF
--- a/drgn_tools/iscsi.py
+++ b/drgn_tools/iscsi.py
@@ -1,0 +1,156 @@
+# Copyright (c) 2024, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
+"""
+Helper for iscsi
+"""
+import argparse
+from typing import Iterator
+
+from drgn import Object
+from drgn import Program
+from drgn.helpers.common import escape_ascii_string
+from drgn.helpers.common.type import enum_type_to_class
+from drgn.helpers.linux.list import list_for_each_entry
+
+from drgn_tools.corelens import CorelensModule
+from drgn_tools.module import ensure_debuginfo
+from drgn_tools.scsi import for_each_scsi_host
+from drgn_tools.scsi import for_each_scsi_host_device
+from drgn_tools.scsi import host_module_name
+from drgn_tools.scsi import scsi_device_name
+from drgn_tools.table import print_dictionary
+from drgn_tools.util import enum_name_get
+
+ISCSI_SESSION_STATES = ["LOGGED_IN", "FAILED", "FREE"]
+
+
+def for_each_iscsi_host(prog: Program) -> Iterator[Object]:
+    """
+    Iterate through all scsi host devices and returns an
+    iterator of hosts backed by iscsi drivers.
+    :returns: an iterator of ``struct Scsi_Host *``
+    """
+
+    # additional iscsi drivers
+    ISCSI_MODS = [
+        "qla4xxx",
+        "bnx2i",
+        "cxgb3i",
+        "cxgb4i",
+        "qedi",
+        "iscsi_tcp",
+        "be2iscsi",
+    ]
+    for shost in for_each_scsi_host(prog):
+        mod_name = host_module_name(shost).lower()
+        if mod_name in ISCSI_MODS:
+            yield shost
+
+
+def for_each_iscsi_session(prog: Program) -> Iterator[Object]:
+    """
+    Iterate through all iscsi_cls_session and gets the associated iscsi_session.
+    :returns: an iterator of ``struct iscsi_session *``
+
+    """
+    for cls_session in list_for_each_entry(
+        "struct iscsi_cls_session", prog["sesslist"].address_of_(), "sess_list"
+    ):
+        yield Object(
+            prog,
+            prog.type("struct iscsi_session", filename="libiscsi.h"),
+            address=cls_session.dd_data,
+        )
+
+
+def print_iscsi_sessions(prog: Program) -> None:
+    """
+    Dump iscsi sessions.
+
+    """
+    msg = ensure_debuginfo(prog, ["libiscsi", "scsi_transport_iscsi"])
+    if msg:
+        print(msg)
+        return
+
+    output = {}
+
+    for session in reversed(list(for_each_iscsi_session(prog))):
+        print("**********")
+        conn = session.leadconn
+
+        persistent_address = escape_ascii_string(
+            conn.persistent_address.string_()
+        )
+        persistent_port = int(conn.persistent_port)
+        output[
+            "Scsi_Host"
+        ] = f"host{session.host.host_no.value_()} ({hex(session.host.value_())})"
+        output["Session"] = hex(session.address_of_())
+        output["SID"] = str(int(session.cls_session.sid))
+        output["Persistent Portal"] = f"{persistent_address}:{persistent_port}"
+        output["Iface Name"] = escape_ascii_string(session.ifacename.string_())
+        output["Session State"] = ISCSI_SESSION_STATES[
+            session.cls_session.state
+        ]
+        connstate = enum_type_to_class(
+            prog.type("enum iscsi_connection_state"), "connstate"
+        )
+        output["Connection State"] = str(
+            enum_name_get(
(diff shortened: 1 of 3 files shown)


drgn_tools/iscsi.py
```diff
@@ -0,0 +1,132 @@
...
+        prog,
+        "struct class",
+        address=prog["iscsi_host_class"].address_of_().value_(),
+    ).p.klist_devices.k_list.address_of_()
+
+    if class_in_private:
+        for device_private in list_for_each_entry(
+            "struct device_private", devices, "knode_class.n_node"
+        ):
+            yield device_private.device
```
biger410, line 43: In the code path of `class_in_private` is not null, it never checks anything for iscsi, why will it be iscsi device?

drgn_tools/iscsi.py
```diff
@@ -60,7 +65,7 @@ def for_each_iscsi_host_device(prog: Program) -> Iterator[Object]:
             yield dev
 
 
-def for_each_iscsi_host(prog: Program) -> Iterator[Object]:
+def for_each_scsi_host(prog: Program) -> Iterator[Object]:
```
biger410, line 68: `for_each_scsi_host` is already defined in scsi.py, just import it.

drgn_tools/iscsi.py
```diff
@@ -77,28 +82,34 @@ def for_each_iscsi_session(prog: Program) -> Iterator[Object]:
...
+    iscsi_sw_tcp_host_addr = (
+        Scsi_Host.value_()
+        + sizeof(prog.type("struct Scsi_Host"))
+        + sizeof(prog.type("struct iscsi_host"))
+    )
+    iscsi_sw_tcp_host = Object(
+        prog, "struct iscsi_sw_tcp_host", address=iscsi_sw_tcp_host_addr
+    )
 
+    return iscsi_sw_tcp_host.session
```
biger410, line 103: We should do some precheck to make sure this is sw tcp iscsi, if it's not, then leave a message saying it's not supported, otherwise the script will just crash for other iscsi type.

drgn_tools/iscsi.py
```diff
@@ -77,28 +82,34 @@ def for_each_iscsi_session(prog: Program) -> Iterator[Object]:
     :returns: a iterator of ``struct iscsi_session *``
 
     """
-    for h in for_each_iscsi_host(prog):
-        yield __get_iscsi_session(h)
+    for h in for_each_scsi_host(prog):
+        yield __get_iscsi_session(prog, h)
```
biger410, line 86: The system could contain other scsi disk like virtio-scsi, sas or fc disk other than iscsi, here it is iterating scsi host, it will include all other types as well, we should provide a new helper `for_each_iscsi_host` which used `for_each_scsi_host` and then just return the host which is iscsi type.

drgn_tools/iscsi.py
```diff
@@ -126,7 +137,49 @@ def print_iscsi_hosts(prog: Program) -> None:
...
+        output["Session"] = hex(session.value_())
+        output["SID"] = str(int(session.cls_session.sid))
+        output["Persistent Portal"] = f"{persistent_address}:{persistent_port}"
+        output["Iface Name"] = escape_ascii_string(session.ifacename.string_())
+        output["Initiatorname"] = escape_ascii_string(
+            session.initiatorname.string_()
+        )
+        output["Targetname"] = escape_ascii_string(
+            session.targetname.string_()
+        )
```
biger410, line 171: the session and connection status are required. Also please dump all the scsi device attached to this session and disk status as well.

biger410 (commented):
Please reorganize the patch, the second commit should be removed, the 3th to 5th should be merged into one commit. Please create a bug and put the bugdb id in the commit log, also for each commit, please brief in the git commit log what the change is for.

drgn_tools/iscsi.py
```diff
@@ -0,0 +1,213 @@
...
+
+def get_dev_name(prog: Program, sdev: Object) -> str:
+    """
+    Get the device name associated with scsi_device.
+    :return ``str``
+    """
+    rq = sdev.request_queue
+    dev = container_of(rq.kobj.parent, "struct device", "kobj")
+    return dev.kobj.name.string_().decode()
+
```
biger410, line 71: `shost_for_each_device` and `get_dev_name` are generic scsi helpers, should be moved to scsi.py, and also maybe rename it to `for_each_scsi_host_device` and `get_scsi_device_name`?

drgn_tools/iscsi.py
```diff
@@ -0,0 +1,213 @@
...
+def for_each_iscsi_session(prog: Program) -> Iterator[Object]:
+    """
+    Iterates through all iscsi tcp hosts, gets their associated iscsi_sw_tcp_hosts and return a
+    iterator of their sessions.
+    :returns: a iterator of ``struct iscsi_session *``
+
+    """
+    for h in for_each_iscsi_tcp_host(prog):
+        yield __get_iscsi_sw_tcp_host(prog, h).session
+
```
biger410, line 82: As in my following comment, let try to make these helper generic to iscsi drivers, not only to iscsi_tcp driver.

biger410:
One more comment, please change the commit subject for this patch, probably "Add iscsi helpers to dump iscsi session information"

richl9:
I noticed that there were test failures reporting a key error for doing  'prog['sesslist']' for ueknext. I tested it locally with 6.11.0-0 as well but didn't observe such error. Maybe could @brenns10 help take a look? Thanks.

drgn_tools/iscsi.py
```diff
@@ -0,0 +1,149 @@
...
+        print()
+
+
+class IscsiDump(CorelensModule):
+    """
+    Dump iscsi info
+    """
+
+    name = "iscsi"
```
biger410, line 146: Please add `skip_unless_have_kmod = "libiscsi scsi_transport_iscsi"`

drgn_tools/iscsi.py
```diff
@@ -0,0 +1,149 @@
...
+    """
+    for scsi_dev in for_each_scsi_host_device(prog, shost):
+        name = get_scsi_device_name(prog, scsi_dev)
+        yield scsi_dev, name
+
+
+def print_iscsi_report(prog: Program) -> None:
+    """
+    Prints a comprehensive report including iscsi table and session stats. More info can be added later if needed.
+    """
```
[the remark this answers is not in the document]
biger410: Please add below code at the beginning of this function, this will make sure the helper bail out if iscsi is not used when invoking this helper directly from drgn shell while not corelen module.
```
    msg = ensure_debuginfo(prog, ["libiscsi", "scsi_transport_iscsi"])
    if msg:
        print(msg)
        return
```

tests/test_iscsi.py
```diff
@@ -0,0 +1,14 @@
...
+from drgn_tools.module import KernelModule
+
+
+def test_iscsi(prog):
+    # make sure iscsi module is loaded
+    ISCSI_MODS = ["libiscsi", "scsi_transport_iscsi"]
+    for mod in ISCSI_MODS:
+        km = KernelModule.find(prog, mod)
+        if not km:
+            return
```
biger410, line 13: With `ensure_debuginfo` in the beginning of `print_iscsi_report`, above loading debuginfo code is not needed.

biger410:
Also the gitlab CI test failed, please fix it.

drgn_tools/iscsi.py
```diff
@@ -0,0 +1,156 @@
...
+    # additional iscsi drivers
+    ISCSI_MODS_ADD = [
+        "qla4xxx",
+        "libcxgbi",
+        "bnx2i",
+        "cxgb3",
+        "cxgb3i",
+        "cxgb4",
+        "cxgb4i",
+    ]
```
biger410, line 43: `libcxgbi` is a dependency module of `cxgb3i` and `cxgb4i`, we don't need it? And also `cxgb3` and `cxgb4` not a kernel module name?
`qedi` is also iscsi driver, please add.
Please add `iscsi_tcp` and `be2iscsi` in the list, so you don't need check whether iscsi is in the name of driver module below.

biger410 (approved):
Looks good, thank you.

brenns10 (approved):

PR #109, opened 2024-09-30
lockup: Print tasks which have been on-cpu for too long

richl9:
1. add helper for scanning potential lockups on cpus and dumping their info
2. add prio in runq dump

### DIFF
--- a/drgn_tools/lockup.py
+++ b/drgn_tools/lockup.py
@@ -1,0 +1,70 @@
+# Copyright (c) 2024, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
+import argparse
+
+from drgn import Program
+from drgn.helpers.common import escape_ascii_string
+from drgn.helpers.linux.cpumask import for_each_online_cpu
+from drgn.helpers.linux.percpu import per_cpu
+
+from drgn_tools.bt import bt
+from drgn_tools.corelens import CorelensModule
+from drgn_tools.task import task_lastrun2now
+from drgn_tools.util import timestamp_str
+
+
+def scan_lockup(
+    prog: Program, min_run_time_seconds: int = 1, skip_swapper: bool = True
+) -> None:
+    """
+    Scan potential lockups on cpus.
+
+    :param prog: drgn program
+    :param min_run_time_seconds: int
+    :param skip_swapper: bool
+    """
+    nr_processes = 0
+    for cpus in for_each_online_cpu(prog):
+        runqueue = per_cpu(prog["runqueues"], cpus)
+        curr_task_addr = runqueue.curr.value_()
+        curr_task = runqueue.curr[0]
+        comm = escape_ascii_string(curr_task.comm.string_())
+        pid = curr_task.pid.value_()
+        run_time = task_lastrun2now(curr_task)
+        prio = curr_task.prio.value_()
+        if run_time < min_run_time_seconds * 1e9:
+            continue
+        if skip_swapper and comm == f"swapper/{cpus}":
+            continue
+        print(f"CPU {cpus} RUNQUEUE: {runqueue.address_of_().value_():x}")
+        print(
+            f"  PID: {pid:<6d}  TASK: {curr_task_addr:x}  PRIO: {prio}"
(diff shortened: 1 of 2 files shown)


drgn_tools/lockup.py
```diff
@@ -0,0 +1,72 @@
...
+
+    print(
+        f"We found {nr_processes} processes running more than {min_run_time_seconds} seconds."
+    )
+
+
+class LockUp(CorelensModule):
+    """
+    Corelens Module for lockup
+    """
```
brenns10, line 58: This text appears in the `corelens -L` output. It should be only one line, and useful to tell users what it does. EG:
```suggestion
    """Print tasks which have been on-cpu for too long"""
```

drgn_tools/lockup.py
```diff
@@ -0,0 +1,72 @@
...
+
+    name = "lockup"
+
+    def add_args(self, parser: argparse.ArgumentParser) -> None:
+        parser.add_argument(
+            "--time",
+            "-t",
+            type=float,
+            default=0,
```
brenns10, line 67: The default of 0 seems unhelpful, since that prints any running task, right? Can we specify a more useful default?

Does `--time` specify the amount of time that a task has been on-CPU without the opportunity to schedule? Or is it just for the current time slice? If there's a low load, it would be totally reasonable for a task to be on-CPU for a long time, right?

drgn_tools/lockup.py
```diff
@@ -0,0 +1,72 @@
...
+            help="list all the processes that have been running more than <time> seconds",
```
brenns10, line 68: Clarify this: is it the total amount of seconds the task has been running? OR just since the last schedule? Or just since the last `cond_resched()`?

drgn_tools/runq.py
```diff
@@ -65,20 +67,31 @@ def _print_cfs_runq(runqueue: Object) -> None:
...
+        run_time = task_lastrun2now(curr_task)
+        prio = curr_task.prio.value_()
+        if run_time < min_run_time_seconds * 1e9:
+            continue
         print(f"CPU {cpus} RUNQUEUE: {runqueue.address_of_().value_():x}")
         print(
-            f"  CURRENT:   PID: {pid:<6d}  TASK: {runqueue.curr.value_():x}"
+            f"  CURRENT:   PID: {pid:<6d}  TASK: {curr_task_addr:x}  PRIO: {prio}"
             f'  COMMAND: "{comm}"',
+            f"  RUNTIME: {timestamp_str(run_time)}",
```
brenns10, line 94: Adding this string after a comma makes it a second argument to `print()`.

The rest of these strings are specified without commas, which instead performs string concatenation. Could you please get rid of the comma between these two lines?

drgn_tools/runq.py
```diff
@@ -65,20 +67,31 @@ def _print_cfs_runq(runqueue: Object) -> None:
         print("     [no tasks queued]")
 
 
-def run_queue(prog: Program) -> None:
+def run_queue(prog: Program, min_run_time_seconds: int = 0) -> None:
     """
-    Print tasks which are in the RT and CFS runqueues on each CPU
+    Print tasks which are in the RT and CFS runqueues on each CPU. Specify min_run_time_seconds to x to only print
+    processes running more than x seconds. Set skip_swapper = True to not include swapper process in the ouput.
```
brenns10, line 73: There's no `skip_swapper` arg here. Either add it or delete the sentence.

brenns10 (changes requested):
Looks interesting. I think my biggest question is whether the number of seconds on CPU is actually a reliable indicator of some sort of lockup, and whether 0 is a good default for that parameter.

drgn_tools/lockup.py
richl9: Right. Forgot to change it back after testing. Was thinking about default = 2. What do you think?  And --time is for the current time slice.

drgn_tools/lockup.py
richl9: I got them time from task_lastrun2now(), so it's the duration from task last run timestamp to now.

drgn_tools/lockup.py
brenns10: That seems reasonable, we can always adjust it in the future.

brenns10 (approved):

PR #110, opened 2024-10-01
slabinfo: detect circular freelists

brenns10:
We use a set to track the freelist, but we don't actually use it to check whether we've seen the object in the freelist. If an object gets double freed, we'll enter into an infinite loop.

Check this and raise an error. Since we're already detecting invalid freelist pointers as one kind of error, let's add an exception class hierarchy to represent freelist errors, and then we can put this into the slabinfo directly. Then, we can format the information in a table at the end of the slabinfo.

### DIFF
--- a/drgn_tools/slabinfo.py
+++ b/drgn_tools/slabinfo.py
@@ -6,7 +6,6 @@ Helper to view slabinfo data
 import argparse
 from typing import List
 from typing import NamedTuple
-from typing import Set
 from typing import Tuple
 
 from drgn import cast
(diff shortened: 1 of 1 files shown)


brenns10:
Example output:

```
...
ffff99f1d69a9880        232      100133*      304       19       4k  skbuff_head_cache
...
WARNING: freelist corruption was detected. It is likely that a use-after-free or double-free bug occurred.
CACHE                     CPU  ERROR
skbuff_head_cache         0    duplicate freelist entry on cpu 0: ffff99f44bf6c800
skbuff_head_cache         1    duplicate freelist entry on cpu 1: ffff99f22e558c00
```

imran-kn (approved):
Ship it :)

PR #112, opened 2024-10-08
ls: Add support to recursively print dentries of subdirectories

srivathsa729:
Currently, ls module prints the dentries of a given directory. Add support to print the dentries of its subdirectories recursively up to a specified depth.

Signed-off-by: Srivathsa Dara <srivathsa.d.dara@oracle.com>

### DIFF
--- a/drgn_tools/dentry.py
+++ b/drgn_tools/dentry.py
@@ -381,24 +381,47 @@ def __file_type(mode: Object) -> str:
     return "UNKN"
 
 
-def ls(prog: Program, directory: str, count: bool = False) -> None:
+def ls(
+    prog: Program,
+    directory: str,
+    parent_dentry: Optional[Object] = None,
+    depth: int = 0,
+    count: bool = False,
+) -> None:
     """
     Print dentry children, like the ls command
     :param directory: directory to print children of
+    :param parent_dentry: dentry object of the directory to print children of, if not provided the dentry
+                        object will be determined by looking up the directory by name
+    :param depth: If provided, it recursively prints the children of subdirectories up to the specified depth.
     :param count: when true, only print counts (not the full contents)
     """
-    dentries = dentry_for_each_child(path_lookup(prog, directory).dentry)
 
+    subdirs = []
+    dentries = dentry_for_each_child(
+        parent_dentry if parent_dentry else path_lookup(prog, directory).dentry
+    )
+
+    print(f"{directory}:")
     pos = neg = 0
     for i, dentry in enumerate(dentries):
-        path = dentry_path_any_mount(dentry).decode()
+        name = dentry.d_name.name.string_().decode()
         if dentry_is_negative(dentry):
             neg += 1
+            dentry_type = "NEG"
         else:
+            dentry_type = __file_type(int(dentry.d_inode.i_mode))
+            if dentry_type == "DIR":
+                subdirs.append((f"{directory}/{name}", dentry))
             pos += 1
         if not count:
-            print(f"{i:05d} {path}")
-    print(f"{pos} positive, {neg} negative dentries")
+            print(f"\t{i:05d} {dentry_type:4s} {name}")
+    print(f"\t{pos} positive, {neg} negative dentries")
+
+    if depth == 0:
+        return
+    for dir in subdirs:
+        ls(prog, dir[0], dir[1], depth - 1, count)
 
 
 class Ls(CorelensModule):
@@ -411,6 +434,14 @@ class Ls(CorelensModule):
     run_when = "never"
 
     def add_args(self, parser: argparse.ArgumentParser) -> None:
+        def non_negative_int(value):
+            ivalue = int(value)
+            if ivalue < 0:
+                raise argparse.ArgumentTypeError(
+                    f"Invalid depth value: {value}. Depth must be a non-negative integer."
+                )
+            return ivalue
+
         parser.add_argument(
             "directory",
             type=str,
@@ -422,9 +453,16 @@ class Ls(CorelensModule):
             action="store_true",
             help="only print counts, rather than every element",
         )
+        parser.add_argument(
+            "--depth",
+            "-d",
+            type=non_negative_int,
+            default=0,
+            help="Print the dentries of subdirectories recursively up to the specified depth.",
+        )
 
     def run(self, prog: Program, args: argparse.Namespace) -> None:
-        ls(prog, args.directory, count=args.count)
+        ls(prog, args.directory, None, depth=args.depth, count=args.count)
 
 
 class DentryCache(CorelensModule):
--- a/tests/test_dentry.py
+++ b/tests/test_dentry.py
@@ -17,4 +17,4 @@ def test_list_dentries_in_hashtable(prog):
 
 
 def test_ls(prog):
-    dentry.ls(prog, "/")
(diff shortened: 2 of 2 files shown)


drgn_tools/dentry.py
```diff
@@ -411,6 +431,14 @@ class Ls(CorelensModule):
...
 
     def add_args(self, parser: argparse.ArgumentParser) -> None:
+        def non_negative_int(value):
+            ivalue = int(value)
+            if ivalue < 0:
+                raise argparse.ArgumentTypeError(
+                    f"Invalid depth value: {value}. Depth must be a non-negative integer."
+                )
+            return ivalue
+
```
brenns10, line 441: nice :) 👍🏻

drgn_tools/dentry.py
```diff
@@ -381,24 +381,44 @@ def __file_type(mode: Object) -> str:
     return "UNKN"
 
 
-def ls(prog: Program, directory: str, count: bool = False) -> None:
+def ls(
+    prog: Program,
+    directory: str,
+    parent_dentry: Object,
+    depth: int,
```
brenns10, line 388: This makes these two new arguments _required_, which makes the ls function much less convenient to call from the CLI. Let's make these optional:
```suggestion
    parent_dentry: Optional[Object] = None,
    depth: int = 0,
```

drgn_tools/dentry.py
```diff
@@ -381,24 +381,44 @@ def __file_type(mode: Object) -> str:
...
+            dentry_type = "negative"
         else:
+            if stat.S_ISDIR(int(dentry.d_inode.i_mode)):
+                subdirs.append([path, dentry])
             pos += 1
         if not count:
-            print(f"{i:05d} {path}")
-    print(f"{pos} positive, {neg} negative dentries")
+            print(f"\t{i:05d} {path} ({dentry_type})")
+    print(f"\t{pos} positive, {neg} negative dentries")
```
brenns10, line 418: This isn't part of your change, but I think printing the full path is a bit too much for each dentry. For deeply nested filesystem paths that's a waste. Let's just print the dentry name!

So instead of using `dentry_path_any_mount()` we can do:

```
name = dentry.d_name.name.string().decode()
...
print(f" .... {name} ...")
```

And then, for the subdirectories, we can do:

```
subdirs.append((f"{directory}/{name}", dentry))
```

drgn_tools/dentry.py
```diff
@@ -381,24 +381,44 @@ def __file_type(mode: Object) -> str:
...
     pos = neg = 0
     for i, dentry in enumerate(dentries):
+        dentry_type = "positive"
         path = dentry_path_any_mount(dentry).decode()
         if dentry_is_negative(dentry):
             neg += 1
+            dentry_type = "negative"
         else:
+            if stat.S_ISDIR(int(dentry.d_inode.i_mode)):
+                subdirs.append([path, dentry])
```
brenns10, line 414: I'd prefer using a tuple instead of a list, it's less expensive:
```
subdirs.append((path, dentry))
```

drgn_tools/dentry.py
```diff
@@ -381,24 +381,44 @@ def __file_type(mode: Object) -> str:
...
+
+    subdirs = []
+    dentries = dentry_for_each_child(
+        parent_dentry if parent_dentry else path_lookup(prog, directory).dentry
+    )
 
+    print(f"{directory}:")
     pos = neg = 0
     for i, dentry in enumerate(dentries):
+        dentry_type = "positive"
```
brenns10, line 407: For positive dentries, can we also distinguish between files and directories (and maybe sockets? as done by the dentrycache tool).

drgn_tools/dentry.py
```diff
@@ -381,24 +381,44 @@ def __file_type(mode: Object) -> str:
...
+                subdirs.append([path, dentry])
             pos += 1
         if not count:
-            print(f"{i:05d} {path}")
-    print(f"{pos} positive, {neg} negative dentries")
+            print(f"\t{i:05d} {path} ({dentry_type})")
+    print(f"\t{pos} positive, {neg} negative dentries")
+
+    for dir in subdirs:
+        ls(prog, dir[0], dir[1], depth - 1, count)
```
brenns10, line 421: nit: for large directories with many sub-directories, it's a bit of a waste to iterate through this list if `depth == 0`. Can you check that case and terminate early here?

drgn_tools/dentry.py
```diff
@@ -381,24 +381,44 @@ def __file_type(mode: Object) -> str:
...
+    directory: str,
+    parent_dentry: Object,
+    depth: int,
+    count: bool = False,
+) -> None:
     """
     Print dentry children, like the ls command
     :param directory: directory to print children of
     :param count: when true, only print counts (not the full contents)
     """
```
brenns10, line 395: Don't forget to update the docstring

brenns10 (changes requested):

drgn_tools/dentry.py
```diff
@@ -381,24 +381,46 @@ def __file_type(mode: Object) -> str:
...
+            dentry_type = "negative"
         else:
+            dentry_type = __file_type(int(dentry.d_inode.i_mode))
+            if dentry_type == "DIR":
+                subdirs.append((f"{directory}/{name}", dentry))
             pos += 1
         if not count:
-            print(f"{i:05d} {path}")
-    print(f"{pos} positive, {neg} negative dentries")
+            print(f"\t{i:05d} {name} ({dentry_type})")
```
brenns10, line 417: I'm so sorry for only catching this now, since the PR looks great otherwise. I think the output would be much easier to scan and read if we put it in columns like below:
```suggestion
            print(f"\t{i:05d} {dentry_type:4s} {name}")
```
And then please make the `negative` text be `NEG` so that it is similar to the other dentry types.

drgn_tools/dentry.py
```diff
@@ -381,24 +381,46 @@ def __file_type(mode: Object) -> str:
...
+    prog: Program,
+    directory: str,
+    parent_dentry: Optional[Object] = None,
+    depth: int = 0,
+    count: bool = False,
+) -> None:
     """
     Print dentry children, like the ls command
     :param directory: directory to print children of
+    :param parent_dentry: dentry object of the directory to print children of
```
brenns10, line 394: maybe mention that "if not provided, the dentry will be determined by looking up the directory by name"

brenns10 (changes requested):
I'm sorry for not making these requests on the previous review, the PR looks great but I have two small nitpicks. Can you also take a look at the "Developer Workflow For drgn-tools" (check slack) and file the necessary Orabugs for this change too? I don't think this necessarily needs to go back to 1.x stable branch, but if you think it's worth it, that will be a second Orabug and a pull request against `stable/v1` as the document states.

brenns10 (approved):
Great, thanks for bearing with me! This is actually really fun to play around with on my laptop, `corelens /proc/kcore -M ls -d 2 -c ~` gives interesting output to learn about how negative dentries get created on the desktop and where.

srivathsa729:
Np, but I see the pipeline in gitlab failed. Is that expected?

brenns10:
Yes, I checked the pipeline and the failure has to do with a race condition in one of the test libraries - not your changes.

PR #140, opened 2024-12-25
helper and corelens module to show unsubmitted pending works.

imran-kn:
Add helper and corelens module to show unsubmitted pending works.

dealyed_work(s) get their pending bit set but are actually submitted to a workqueue,
upon expiration of corresponding timer(s).
Recently we have found some cases where a delayed work submitted to an already
offlined CPU was never getting executed, because underlying timers were not
 firing in first place. Since the pending bit was set, this gave a notion that
 work item was lost to workqueue subsystem (which was not the case here.)
    
  Add an helper and a corelens module to dump delayed_work(s) whose timer has
  not yet expired. This is off interest for offline CPUs mainly, because ideally
  we should not see any delayed_work timer lying on an offlined CPU. So by default
  the helper and corelens module dump this info for offlined CPUs only

### DIFF
--- a/drgn_tools/workqueue.py
+++ b/drgn_tools/workqueue.py
@@ -684,13 +684,13 @@ class OfflinedDelayedWorksModule(CorelensModule):
     """
     Show delayed works from offlined CPUs.
     Delayed works (with non zero delay), rely on timer-wheel timers for
-    their submission. If these timers don't fire the work does not get
+    their submission. If these timers don't fire, the work does not get
     submitted. So delayed works submitted to an offlined CPU, don't get
     executed even after specified delay because timer-wheel timers on
     offlined CPUs don't get fired in first place.
 
     This corelens module list delayed works on offlined CPUs, so that
-    one can know if a delayed work was left unexececuted, due to the fact
+    one can know if a delayed work was left unexecuted, due to the fact
     that it was submitted on an offlined CPU.
     """
 
--- a/drgn_tools/workqueue.py
+++ b/drgn_tools/workqueue.py
@@ -13,11 +13,17 @@ from typing import Optional
 from typing import Union
 
 from drgn import cast
+from drgn import container_of
+from drgn import FaultError
 from drgn import IntegerLike
 from drgn import NULL
 from drgn import Object
 from drgn import Program
+from drgn import sizeof
 from drgn.helpers.common.format import escape_ascii_string
+from drgn.helpers.linux.bitops import for_each_set_bit
+from drgn.helpers.linux.cpumask import for_each_online_cpu
+from drgn.helpers.linux.cpumask import for_each_possible_cpu
 from drgn.helpers.linux.idr import idr_find
 from drgn.helpers.linux.idr import idr_for_each
 from drgn.helpers.linux.list import hlist_for_each_entry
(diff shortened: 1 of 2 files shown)


biger410 (commented):
The helper looks great. One question about the function name, why is it marked as "UNKNOWN"?

imran-kn:
> The helper looks great. One question about the function name, why is it marked as "UNKNOWN"?

Thanks @biger410 for taking a look. The "UNKNOWN" appears as function name because I had not loaded the module symbols. We try with prog.symbol to get the function name and if prog.symbol returns LookupError, we put "UNKNOWN" in function name

drgn_tools/workqueue.py
```diff
@@ -614,6 +621,74 @@ def find_worker_executing_work(work: Object) -> Object:
     return NULL(prog, "struct worker *")
 
 
+def show_pending_but_unsubmitted_delayed_works(
+    prog: Program, only_offline_cpus: bool = True
+) -> None:
```
biger410, line 626: I know that "pending" is for "WORK_STRUCT_PENDING_BIT" bit and "unsubmitted" is for worker not reaching worker pool, but from user point of view, would either of them mean that the delay_works has not yet expired? How about name the function "show_unexpired_delayed_works"?

drgn_tools/workqueue.py
```diff
@@ -614,6 +621,74 @@ def find_worker_executing_work(work: Object) -> Object:
...
+                            )
+
+        except FaultError:
+            continue
+
+
+class UnsubmittedPendingWorkModule(CorelensModule):
+    """Show pending but unsubmitted works"""
+
+    name = "unsubmitted_pending_works"
```
biger410, line 686: This corelen module is only for unexpired delayed_work from offlined cpu, how about name it as "offlined_delayed_works"? 
Also please enhancement the docstring of it to make sure user know this is for delayed_work from offline cpu and it will never expire until the cpu is offline again, maybe remove "pending" and "unsubmitted" in the string, user may not know what that mean.

imran-kn: I have done these changes. Could you please have a look to see if docstring looks good now

drgn_tools/workqueue.py
imran-kn: Makes sense. I have changed the function name.

drgn_tools/workqueue.py
```diff
@@ -614,6 +621,85 @@ def find_worker_executing_work(work: Object) -> Object:
...
+    Delayed works (with non zero delay), rely on timer-wheel timers for
+    their submission. If these timers don't fire the work does not get
+    submitted. So delayed works submitted to an offlined CPU, don't get
+    executed even after specified delay because timer-wheel timers on
+    offlined CPUs don't get fired in first place.
+
+    This corelens module list delayed works on offlined CPUs, so that
+    one can know if a delayed work was left unexececuted, due to the fact
+    that it was submitted on an offlined CPU.
+    """
```
biger410, line 695: Please add a "," after "fire" in "If these timers don't fire the work does not get". Also fix the typo "unexececuted"

imran-kn: Done

PR #142, opened 2025-01-07
Add more info to iscsi dump info

gulams:
Orabug: 34461976

### DIFF
--- a/drgn_tools/iscsi.py
+++ b/drgn_tools/iscsi.py
@@ -5,6 +5,12 @@ Helper for iscsi
 """
 import argparse
 from typing import Iterator
+from typing import Dict
+from typing import Any
+
+import drgn
+import socket
+from drgn import cast
 
 from drgn import Object
 from drgn import Program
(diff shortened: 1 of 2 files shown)


drgn_tools/iscsi.py
```diff
@@ -21,8 +27,83 @@
...
+# iSCSI Session States
+iscsi_session_states = ["UNKNOWN",
+                        "FREE",
+                        "LOGGED_IN",
+                        "FAILED",
+                        "TERMINATE",
+                        "IN_RECOVERY",
+                        "RECOVERY_FAILED",
+                        "LOGGING_OUT"
+                        ]
```
brenns10, line 49: First: please make all global variables named in `ALL_CAPS`, since it makes reading the code that uses them much easier. This should apply for the rest of the globals here.

Second: These states appear to be an enumeration from `libiscsi.h`. For enums, the best thing to do is rely on the kernel's debuginfo to contain the enumeration constants. Sadly, `libiscsi.h` makes these enumerations anonymous, so it's slightly difficult to get them. Here's an example:

```python
from typing import Dict, Type
from drgn import Program
from drgn.helpers.common import enum_type_to_class

def iscsi_constants(prog: Program) -> Dict[str, Type]:
    return {
        "ISCSI_STATE": enum_type_to_class(
            prog.constant("ISCSI_STATE_FREE").type_,
            "IscsiState",
            prefix="ISCSI_STATE",
        ),
        ...
    }
```

You can do this for each enum: choose the most common variant (the one whose name is least likely to change), and then create the Python class with [enum_type_to_class()](https://drgn.readthedocs.io/en/latest/helpers.html#drgn.helpers.common.type.enum_type_to_class).

Once you have the python enum class corresponding to it, it will be quite easy to just lookup the name. For instance:

```python
state = 5  # in reality, this would come from the program
consts = iscsi_constants(prog)
value = consts(state)
print(str(value))
```

That goes for all these lists of constants below - look them up from the program and create them dynamically. That way if any constants change in any kernel version, you'll get the right values.

drgn_tools/iscsi.py
```diff
@@ -73,76 +241,15 @@ def print_iscsi_sessions(prog: Program) -> None:
...
-                sdev_state,
-                scsi_dev.sdev_state,
-                "UNKNOWN",
-            )
-            print("  Attached iscsi disk: {}  State: {}".format(name, state))
-
-        print()
-
+            iscsi_disks_info = get_iscsi_disks_info(scsi_dev, prog)
+            print_iscsi_info(iscsi_disks_info)
```
brenns10, line 252: Could this be a table instead? It would probably make it easier to read. See the `drgn_tools/table.py`.

drgn_tools/iscsi.py
```diff
@@ -21,8 +27,83 @@
...
+        if "Target IQN" in line:
+            print(f"{line}: {info[line]}")
+        elif line == "Current Portal" or line == "Persistent Portal":
+            print(f"\t{line}: {info[line]}")
+        elif any(line in s for s in sub_headings):
+            print(f"\t\t"+"*" * (len(line) + 1))
+            print(f"\t\t{line}: {info[line]}")
+            print(f"\t\t"+"*" * (len(line) + 1))
+        else:
+            print(f"\t\t{line}: {info[line]}")
```
brenns10, line 106: I do like seeing a separation between helpers that _get_ information, and those that _print_ information. That makes it easier for others to write new code based on your helpers. So I do like what you're trying to do here.

But I'm not certain that it's really helping to have a `print_iscsi_info()` function and a `get_iscsi_info()` here, for a few reasons.

1. The dictionary returned by `get_iscsi_info()` is just a bunch of strings. It seems to me that if a developer wanted to get the timeout associated with an `iscsi_session`, they would want it as a numeric value, not a string that they would need to interpret. So if you're going to make a `get_iscsi_info()` function, it should return data in a form that a _developer_ would like, not in the form that you want to print it.
2. The use of blank "header" keys is a confusing design choice, that can easily go out-of-date if somebody updates the `get_iscsi_info()` function, but forgets to update the list of `sub_headings`.

I think there are two ways to go about this. Either change `get_iscsi_info()` to just be `print_iscsi_info()`, so that it directly prints the data without sticking it into a dict first, or else change `get_iscsi_info()` to return the data in a developer-friendly format (e.g. a dataclass or named tuple) and then format it dircetly in the `print_iscsi_info()` function.

For an example of the latter, see ` get_kmem_cache_slub_info()` in `drgn_tools/slabinfo.py`. It returns a named tuple that is later formatted (as part of a table).

drgn_tools/iscsi.py
```diff
@@ -62,8 +143,95 @@ def for_each_iscsi_session(prog: Program) -> Iterator[Object]:
...
+                                        (current_address >> 24) & 0xFF]))
+            current_port = iscsi_sock.__sk_common.skc_dport
+            current_port = prog.read_u16(current_port.address_of_())
+            current_port = socket.ntohs(current_port)
+            local_address = iscsi_sock.__sk_common.skc_rcv_saddr
+            local_address = prog.read_u32(local_address.address_of_())
+            local_ip = socket.inet_ntoa(bytes([(local_address) & 0xFF,
+                                        (local_address >> 8) & 0xFF,
+                                        (local_address >> 16) & 0xFF,
+                                        (local_address >> 24) & 0xFF]))
```
brenns10, line 175: For IP addresses, you should use `get_ipv4()` which is defined in `drgn_tools/crash_net.py`, see its use there. The kernel always stores the IP addresses in network byte order, so as long as you use that function, you don't need to do byte swapping.

For the port, only do the byte swap if the program is little endian. (which is true for all the architectures we use, but we really should do it correctly). For example, see these lines in the `crash_net.py` file:
```
            dport = int(skc.skc_dport)
            if prog.platform.flags & PlatformFlags.IS_LITTLE_ENDIAN:
                dport = socketlib.ntohs(dport)
```

brenns10 (changes requested):
I've left specific feedback below, but two pieces of general feedback regarding the whole patch:

1. The second commit in this branch does not have a `Signed-off-by`. This is why the bot has marked this with a "needs OCA". I think that second commit probably should just be folded into the first one. So please submit this as a single commit, making sure you have a sign-off in it.
2. The code here is not formatted according to the guidelines. You can see the [pre-commit output here](https://github.com/oracle-samples/drgn-tools/actions/runs/12649781748/job/35246910346?pr=142). Please be sure to set up pre-commit in your development environment and use it for future submissions.

Both of these are covered in the developer workflow doc, reach out to me if you need more info.

biger410:
Discussed with Gulam, he will address the legal requirement and open a new PR with all above @brenns10 's feedback addressed.

PR #146, opened 2025-01-17
Print KVM memslots related information

SiddhiK29:


drgn_tools/kvm.py
```diff
@@ -183,6 +182,70 @@ def print_vcpu_list(prog: Program) -> None:
...
+            "PAGES",
+            "ARCH",
+            "USER_ADDR",
+            "FLAGS",
+        ]
+    ]
+
+    for vm in vm_list:
+        for i in range(len(vm.memslots)):
+            kvm_memslots = hex(vm.memslots[i])
```
brenns10, line 202: Each time you do `vm.memslots[i]`, drgn needs to create a new Object to represent it. Unfortunately, Python can't optimize these things like a C compiler could. So it would be much better to write the loop like this:
```
for memslot in vm.memslots:
    kvm_memslots = hex(memslot)
    ...
```
That way you've created the `memslot` object once, and you can use it in place of `vm.memslots[i]` for the body of the loop below. The variable name obviously could be changed - I'm not familiar with this, and my name may not be a very good one.

drgn_tools/kvm.py
```diff
@@ -183,6 +182,70 @@ def print_vcpu_list(prog: Program) -> None:
...
+            if has_member(vm.memslots[i], "memslots"):
+                for j in range(vm.memslots[i].used_slots.value_()):
```
brenns10, line 204: Similarly, it would be slightly better to do:
```suggestion
                for j in range(vm.memslots[i].used_slots.value_()):
                    mm = memslots[j]
```
And then use `mm` (or a better variable name) in place of `vm.memslots[i].memslots[j]` below.

brenns10 (changes requested):
I have a few smaller comments. But I admit I'm not an expert, so I'll ask Joe to take a look at this as well.

drgn_tools/kvm.py
SiddhiK29: Thank you for the suggestion Stephen. I have made the changes accordingly.

drgn_tools/kvm.py
```diff
@@ -183,6 +182,66 @@ def print_vcpu_list(prog: Program) -> None:
...
+            "KVM_MEMORY_SLOT",
+            "BASE_GFN",
+            "PAGES",
+            "ARCH",
+            "USER_ADDR",
+            "FLAGS",
+        ]
+    ]
+    for vm in vm_list:
+        for memslot in vm.memslots:
```
joejin00, line 200: Could you add a comment to indicate which kernel version is for this branch, also comment following else branch

joejin00:
Please add below to the output:
1. total pages for each vm
2. translate the flag to string.

drgn_tools/kvm.py
SiddhiK29: Sure, Joe.

drgn_tools/kvm.py
```diff
@@ -182,6 +191,82 @@ def print_vcpu_list(prog: Program) -> None:
...
+            else:
+                # Starting from UEK7-U3
+                for vcpu in for_each_vcpu(vm):
+                    mmslot = vcpu.last_used_slot
+                    mm = hex(mmslot.address_of_())
+                    gfn = mmslot.base_gfn.value_()
+                    pages = mmslot.npages.value_()
+                    arch = hex(mmslot.arch.address_of_())
+                    usr_addr = hex(mmslot.userspace_addr.value_())
+                    flags = mmslot.flags.value_()
```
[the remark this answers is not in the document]
SiddhiK29: yes, Joe.

joejin00:
Data structure in kernel and output is good to me.
Reviewed-By: Joe Jin <joe.jin@oracle.com>

brenns10 (approved):

PR #150, opened 2025-01-31
kernfs,memcg: remove merged slab cache check.

imran-kn:
Remove slab_cache_is_merged check from kernfs_node iterator. When we iterate through kernfs_node slab-cache, we are looking for kernfs_node objects of specific types, for example cgroup, kobjects etc. This means we do some checks on obtained kernfs_node objects.
So rather than giving up in case of merged slab caches, we can still use the above mentioned checks to ensure that we are using correct kernfs_node object.

### DIFF
--- a/drgn_tools/kernfs_memcg.py
+++ b/drgn_tools/kernfs_memcg.py
@@ -24,7 +24,6 @@ from drgn.helpers.linux import inode_path
 from drgn.helpers.linux import kernfs_path
 from drgn.helpers.linux import PageSlab
 from drgn.helpers.linux import slab_cache_for_each_allocated_object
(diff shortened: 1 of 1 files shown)


biger410 (commented):
The patch looks good, please create a bug to track it.

brenns10:
Agreed with @biger410, the patch is good to me, just file a bug and include it in the patch. OR we can use the bug we filed for the original kernfs_memcg feature. That would be fine to me - this fixes/improves something which isn't released yet, so I think it's fair.

imran-kn:
Thanks @brenns10 and @biger410 for reviewing this. I have used bug of original kernfs_memcg feature, to track this one too.

biger410 (approved):

PR #151, opened 2025-02-03
Add "mkdist.py" for creating development distributions

brenns10:


brenns10:
note: this still needs a bug to be filed, consider it a draft for now

PR #152, opened 2025-02-07
Deadlock detection

Pranav-Agarwal0612:
Deadlock Module
	Created a dependency graph between Objects.
	Tested with mutex cycle detection.

Lock Module
	Changed scan_mutex_lock to return its findings for graph creating purpose

### DIFF
--- a/drgn_tools/deadlock.py
+++ b/drgn_tools/deadlock.py
@@ -1,6 +1,5 @@
 # Copyright (c) 2025, Oracle and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
-import argparse
 from typing import Dict
 from typing import List
 from typing import Optional
@@ -9,11 +8,7 @@ from typing import Tuple
 from typing import Union
 
 from drgn import Object
-from drgn import Program
 from drgn import TypeKind
-
-from drgn_tools.corelens import CorelensModule
-from drgn_tools.lock import get_mutex_lock_info
 
 
 class DependencyGraph:
@@ -30,9 +25,9 @@ class DependencyGraph:
                 )
 
             object_node = cls.get_node(
-                name=type_.typename(), address=object.address_
+                name=type_.type_name(), address=object.address_
             )
-            if not object_node.object:
+            if object_node.object is None:
                 object_node.object = object
             return object_node
 
@@ -46,7 +41,7 @@ class DependencyGraph:
             self.address: int = address
             self.depends_on: List[DependencyGraph.Node] = []
             self.blocked_nodes: List[DependencyGraph.Node] = []
-            self.object: Object = None
+            self.object: Optional[Object] = None
 
         def __hash__(self):
             # Using name (example : "struct task_struct") and address as unique
@@ -58,7 +53,7 @@ class DependencyGraph:
             hash_value = hash((name, address))
             if hash_value in DependencyGraph.node_map:
                 return DependencyGraph.node_map[hash_value]
-            return cls(name, address)
+            return cls(name=name, address=address)
 
     node_map: Dict[int, Node] = dict()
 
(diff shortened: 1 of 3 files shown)


drgn_tools/deadlock.py
```diff
@@ -0,0 +1,109 @@
...
+
+            if key not in self.NodeSet:
+                self.NodeSet[key] = self.Node(dependent_object)
+            
+            dependent_node = self.NodeSet[key]
+            dependent_node.depends_on.append(blocking_node)
+
+
+
+    def detect_cycle(self):
```
brenns10, line 58: This needs a type annotation.

brenns10, line 58: This needs a type annotation.

drgn_tools/deadlock.py
```diff
@@ -0,0 +1,109 @@
...
+                dfs(neighbor)
+
+            path.pop() 
+
+        # Run DFS for all nodes in the graph
+        for node in self.NodeSet.values():
+            if node not in visited:  # Only visit unvisited nodes
+                dfs(node)
+
+        return cycles
```
brenns10, line 88: I think the problem we want to solve is larger than simply running DFS on each node. Suppose we have the following graph:

```
A -> D
B -> D
C -> D
D -> C
E -> F
```

There are two things we care about.

1. What is the minimal cycle present? In this case, it is `C <-> D`.
2. What is the maximal set of of elements whose dependencies include the cycle? In this case, it is A, B, C, D.

I think we want to return both pieces of information, and we don't want any duplication. That way, we would be able to report to the user the root cause of the deadlock (that's the first thing), and we'd also be able to go through the second item and report all the tasks or CPUs or whatever are impacted by the cycle, without needing to _spell out_ the entire path for each of them.

drgn_tools/lock.py
```diff
@@ -124,7 +124,7 @@ def scan_mutex_lock(
     stack: bool,
     time: Optional[int] = None,
     pid: Optional[int] = None,
-) -> None:
+) -> List[Tuple[Object, List[Object]]]:
```
brenns10, line 127: Instead of tracking and returning the dependencies, maybe just include an optional argument which is the graph. If the argument is present, then the function would add edges to the graph as it encounters them.

drgn_tools/deadlock.py
Pranav-Agarwal0612: This can easily be solved by changing A `depends_on` B with B `blocks` A. So when we find the cycle, we can also print out the objects/nodes blocked on the nodes present in the cycle. There might be future benefits of keeping both `depends_on` and `blocks`, but I am not sure right now.

drgn_tools/deadlock.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+        def __hash__(self):
+            # Using name (example : struct task_struct") and address as unique 
+            # Will be useful in case of Objects which don't exist in memory (example : CPU) - can set address=0 and name='CPU10'
+            return hash((self.name, self.address))
+        
+        
+    def __init__(self):
+        self.NodeSet: dict[tuple, self.Node] = dict()
```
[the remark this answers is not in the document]
Pranav-Agarwal0612: Fixed

drgn_tools/deadlock.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+    def insert(self, dependency: tuple) -> None:
+        blocking_object: Object = dependency[0]
+        dependent_objects: list[Object] = dependency[1]
+
+        if not blocking_object:
+            return 
+        
+        key = (blocking_object.type_.type_name(), blocking_object.value_())
```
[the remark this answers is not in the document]
Pranav-Agarwal0612: Fixed

drgn_tools/deadlock.py
Pranav-Agarwal0612: Fixed

drgn_tools/deadlock.py
```diff
@@ -0,0 +1,109 @@
...
+
+from drgn_tools.corelens import CorelensModule
+from drgn_tools.lock import scan_mutex_lock
+
+class DependencyGraph:
+
+    class Node:
+        def __init__(self, object: Object) :
+            self.name: str = object.type_.type_name() # TypeName of Object inside Node
+            self.address: int = object.value_() # Memory address of Object in host RAM
```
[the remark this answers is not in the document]
brenns10: Sorry for missing this - if there are, we can add support on a case-by-case basis. I think the vast majority of objects are going to be pointers to structures.

brenns10 (approved):
I think this looks like a great start, thank you!

PR #155, opened 2025-03-26
Add corelens plugin to sosreport

apalakun:
Orabug: Jira-LSE-524
Add corelens plugin to sosreport

### DIFF
--- a/buildrpm/python-drgn-tools.spec
+++ b/buildrpm/python-drgn-tools.spec
@@ -104,6 +104,7 @@ echo '__version__ = "%{version}+%{release}"' > drgn_tools/_version.py
 %py3_install
 gzip man/corelens.1
 install -m644 -D man/corelens.1.gz %{buildroot}%{_mandir}/man1/corelens.1.gz
+install -m644 -D extras/corelens.py %{buildroot}%{python3_sitelib}/sos/report/plugins/
 
 # The DRGN script is an interactive CLI which is convenient for developers,
 # but should not be part of general users' PATH. If necessary, it can be invoked
@@ -116,6 +117,7 @@ rm %{buildroot}/usr/bin/DRGN
 %{python3_sitelib}/drgn_tools/*
 /usr/bin/corelens
 %{_mandir}/man1/corelens.1.gz
+%{buildroot}%{python3_sitelib}/sos/report/plugins/corelens.py
 
 %if %{with_python311}
 %files -n python3.11-drgn-tools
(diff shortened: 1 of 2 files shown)


buildrpm/python-drgn-tools.spec
```diff
@@ -116,6 +117,7 @@ rm %{buildroot}/usr/bin/DRGN
 %{python3_sitelib}/drgn_tools/*
 /usr/bin/corelens
 %{_mandir}/man1/corelens.1.gz
+%{buildroot}%{python3_sitelib}/sos/report/plugins/
```
brenns10, line 120: This indicates that this RPM owns the directory and all its contents. Really we only want to own the file. Can you change this to specify the specific file?

buildrpm/python-drgn-tools.spec
```diff
@@ -21,7 +21,7 @@
 
 Name:           python-drgn-tools
 Version:        2.0.0
-Release:        1%{?dist}
```
brenns10, line 24: Please don't change the version or release, I'll handle that for new releases.

extras/corelens.py
```diff
@@ -0,0 +1,140 @@
+# Copyright (c) 2024, Oracle and/or its affiliates.
```
brenns10, line 1: s/2024/2025

extras/corelens.py
```diff
@@ -0,0 +1,140 @@
...
+                error_logs.append("No recent vmcore directories found.")
+                self.add_string_as_file("\n".join(error_logs), filename="corelens")
+                return
+
+            archive_tmp_dir = self.archive.get_tmp_dir()
+            corelens_output_path = Path(archive_tmp_dir) / "sos_commands/corelens"
+            corelens_output_path.mkdir(parents=True, exist_ok=True)
+
+            kernel_version = subprocess.check_output("uname -r", shell=True, universal_newlines=True).strip()
+            vmlinux_ctf_path = f"/usr/lib/modules/{kernel_version}/kernel/vmlinux.ctfa"
```
brenns10, line 109: You're saying here that the kernel version for this vmcore must be the same as the current kernel. This may not be true. 

I'd recommend not specifying the `--ctf` parameters for corelens. It will determine the kernel version automatically and select CTF (or DWARF) if available.

extras/corelens.py
```diff
@@ -0,0 +1,140 @@
...
+            error_logs.append(f"Unexpected error: {e}")
+
+        if error_logs:
+            self.add_string_as_file("\n".join(error_logs), filename="corelens")
+
+    def setup(self):
+        days = self.get_option('task-days')
+        corelens_binary = Path("/usr/bin/corelens")
+        ctf_path = Path(f"/usr/lib/modules/{subprocess.check_output('uname -r', shell=True).decode('utf-8').strip()}/kernel/vmlinux.ctfa")
+        if corelens_binary.exists() and ctf_path.exists() and self.check_version():
```
brenns10, line 129: Again, I'm not certain why this is here. There's no reason to assume that the vmcore is from the same kernel version as the current kernel. Let corelens determine that.

brenns10 (changes requested):

brenns10:
Closing based on the above, I'll open a new PR and request your review.

PR #159, opened 2025-05-02
lockup: detect the blocker for process hang in RCU grace period

richl9:
Add a new feature in lockup to look for potential blocker (hogging a CPU for too long and RCU cannot go through) for process hang in RCU grace period

### DIFF
--- a/drgn_tools/lockup.py
+++ b/drgn_tools/lockup.py
@@ -1,14 +1,18 @@
-# Copyright (c) 2024, Oracle and/or its affiliates.
+# Copyright (c) 2025, Oracle and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
 import argparse
+import typing
 
+import drgn
 from drgn import Program
 from drgn.helpers.common import escape_ascii_string
 from drgn.helpers.linux.cpumask import for_each_online_cpu
 from drgn.helpers.linux.percpu import per_cpu
 
 from drgn_tools.bt import bt
+from drgn_tools.bt import bt_has_any
 from drgn_tools.corelens import CorelensModule
+from drgn_tools.table import print_table
 from drgn_tools.task import task_lastrun2now
 from drgn_tools.util import timestamp_str
 
@@ -17,7 +21,7 @@ def scan_lockup(
     prog: Program, min_run_time_seconds: int = 1, skip_swapper: bool = True
 ) -> None:
     """
-    Scan potential lockups on cpus.
+    Scan potential lockups on cpus and tasks waiting for RCU.
 
     :param prog: drgn program
     :param min_run_time_seconds: int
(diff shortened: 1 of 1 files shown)


drgn_tools/lockup.py
```diff
@@ -1,26 +1,31 @@
...
+from drgn_tools.bt import bt_has
 from drgn_tools.corelens import CorelensModule
+from drgn_tools.task import task_cpu
 from drgn_tools.task import task_lastrun2now
 from drgn_tools.util import timestamp_str
 
 
 def scan_lockup(
-    prog: Program, min_run_time_seconds: int = 1, skip_swapper: bool = True
+    prog: Program, min_run_time_seconds: int = 1, min_rcu_time_seconds: int = 20, skip_swapper: bool = True
```
biger410, line 21: I don't see we need this new `min_rcu_time_seconds`, 20s is too long for the threshhold, we can reuse `min_run_time_seconds`

drgn_tools/lockup.py
```diff
@@ -47,13 +52,32 @@ def scan_lockup(
...
+                print(f"We found CPU {cpu} waiting on RCU grace period stuck for {run_time} seconds.")
+                break
+
+        print(f"We found {nr_processes} processes running more than {min_run_time_seconds} seconds as above, which are likely RCU grace period blockers.")
+
+    else:
+        print(f"We found {nr_processes} processes running more than {min_run_time_seconds} seconds")
+
+
+def get_rcu_tasks(prog: Program) -> typing.List[typing.Tuple[drgn.Object, drgn.StackFrame]]:
```
biger410, line 70: can we rename this `tasks_waiting_rcu_gp`?
Also It will not be only "percpu_ref_switch_to_atomic_sync", there will be other functions waiting rcu gp as well, to make future adding easier, can you put the function in a table/list, so we just need add a new entry in the table for other functions in future?

drgn_tools/lockup.py
```diff
@@ -47,13 +52,32 @@ def scan_lockup(
...
+            run_time = int(task_lastrun2now(t) / 1e9)
+            cpu = task_cpu(t)
+            if run_time > min_rcu_time_seconds:
+                print(f"We found CPU {cpu} waiting on RCU grace period stuck for {run_time} seconds.")
+                break
+
+        print(f"We found {nr_processes} processes running more than {min_run_time_seconds} seconds as above, which are likely RCU grace period blockers.")
+
+    else:
+        print(f"We found {nr_processes} processes running more than {min_run_time_seconds} seconds")
```
biger410, line 67: I think we don't need dump the original lockup message conditionally, just let it as it and make a new function to dump those tasks waiting rcu gp and invoked it from `scan_lockup`, it should include process name, pid, pending time, format the output as a table output.

biger410 (commented):
Thanks Richard for creating this. I put some comments in the commit and also looks like ci test is failing, please address that as well.

richl9:
> Thanks Richard for creating this. I put some comments in the commit and also looks like ci test is failing, please address that as well.

Thanks a lot for the comments. I have addressed them accordingly. As for the failed ci test, I think it just simply timed-out (as we are searching particular frame names across all stack traces).

drgn_tools/lockup.py
```diff
@@ -47,13 +51,52 @@ def scan_lockup(
         print()
         nr_processes += 1
 
-    print(
-        f"We found {nr_processes} processes running more than {min_run_time_seconds} seconds."
-    )
+        print(f"We found {nr_processes} processes running more than {min_run_time_seconds} seconds")
```
biger410, line 54: The indentation seemed not right?

drgn_tools/lockup.py
```diff
@@ -47,13 +51,52 @@ def scan_lockup(
...
+
+    dump_tasks_waiting_rcu_gp(prog, min_run_time_seconds)
+
+
+def tasks_waiting_rcu_gp(prog: Program) -> typing.List[typing.Tuple[drgn.Object, drgn.StackFrame]]:
+    """
+    Detects any RCU grace period going on and gets the tasks
```
biger410, line 61: "Detects tasks waiting RCU grace period"

drgn_tools/lockup.py
```diff
@@ -47,13 +51,52 @@ def scan_lockup(
...
+
+    :param prog: drgn program
+    :param min_run_time_seconds: int
+    """
+    tasks_waiting = tasks_waiting_rcu_gp(prog)
+    output = [["TASK", "NAME", "PID", "PENDING_TIME (SECONDS)"]]
+    tasks_pids = set()  # remove duplicates
+    if tasks_waiting:
+        for t, _ in tasks_waiting:
+            run_time = int(task_lastrun2now(t) / 1e9)
```
biger410, line 81: Let not convert to second, just use `timestamp_str()`. Also the process is actually in D status, so the variable name "run_time" is not good.

drgn_tools/lockup.py
```diff
@@ -47,13 +51,52 @@ def scan_lockup(
...
+                    [
+                        hex(t.value_()),
+                        escape_ascii_string(t.comm.string_()),
+                        pid,
+                        str(run_time)
+                    ]
+                )
+                tasks_pids.add(pid)
+        print()
+        print("We found below tasks waiting for rcu grace period:")
```
biger410, line 94: `print(f"We found below tasks waiting for rcu grace period over {min_run_time_seconds} seconds:")`

drgn_tools/lockup.py
richl9: Sorry for the oversight. Thanks for pointing out!

biger410 (approved):
Looks good.

PR #160, opened 2025-05-13
Some small fixes

brenns10:


brenns10:
This ended up being more than a few smaller fixes. It turned out that the newest kernel-ueknext RPM was released, and again the debuginfo didn't get posted. I'm working on getting that published, but in the meantime, I've added the ability for us to add this `allow-missing-latest` label to a pull request, so that we'll fall back to using the previous UEK-next release if necessary. That's the bulk of the code in here.

biger410 (approved):

PR #161, opened 2025-05-14
A fix for pre-commit and Linux 6.14

brenns10:


brenns10:
A UEK8 new release just dropped within the last 24 hours. I think the debuginfo is just a bit slow to be uploaded. So I applied allow-missing-latest for that. But the UEK-next tests will still run on 6.14.

brenns10:
Tests pass on vmcores as well. This will resolve a test failure and allow @vashanmu's branch to be reviewed and tested.

PR #162, opened 2025-05-19
Helper to dump ftrace buffer

richl9:


richl9:
wip: fixing some .dat formatting issue

.gitignore
```diff
@@ -13,3 +13,4 @@ drgntools-fio.dat
 testdata
 test.log
 vmcore.xml
+.pre-commit-config.yaml
```
brenns10, line 16: please don't add this to the gitignore

.pre-commit-config.yaml
```diff
@@ -2,7 +2,7 @@
 # See https://pre-commit.com/hooks.html for more hooks
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.4.0
+    rev: v5.0.0
```
brenns10, line 5: I have no problem with updating the pre-commit hooks in general. I'll never turn down maintenance help :)

But usually, changing the pinned version of the pre-commit hooks results in new failures or style updates. So after updating this file, it's important to run `pre-commit run --all-files` and then fix everything there. These are the failures that are popping up in the CI lint runs.

Usually I do this in a separate commit to highlight that every code change was made due to the pre-commit hooks, and not due to any other logical reason.

That said... was this intentional? If not, you can also drop the changes to this file and just focus on your code.

richl9:
Hi @brenns10 , thanks for the comment. This is still wip. It will role out soon:)

brenns10:
Got it, I've marked it as a draft, you can change it back once you're ready for review.

PR #166, opened 2025-06-05
debuginfo: Find vmlinux at other locations

Asphaltt:
On some distributions (e.g., Ubuntu), vmlinux may not reside in the modules directory. Search common locations for vmlinux, as referenced in: https://github.com/anakryiko/retsnoop/blob/389e2c9ddfc686ee048b063ce0c17b94b55398d2/src/retsnoop.c#L59-L67

### DIFF
--- a/drgn_tools/debuginfo.py
+++ b/drgn_tools/debuginfo.py
@@ -345,7 +345,9 @@ def _get_debuginfo_paths(
     return paths
 
 
-def _find_debuginfo(paths: List[Path], mod: str) -> Optional[Path]:
+def _find_debuginfo(
+    prog_or_release: Union["Program", str], paths: List[Path], mod: str
+) -> Optional[Path]:
     replace_pat = re.compile(r"[_-]")
     for search_dir in paths:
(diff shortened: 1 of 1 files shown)


brenns10:
Hi @Asphaltt, I'm actually gearing up to completely revamp this section of drgn-tools. Since drgn 0.0.31, drgn's Module API does a really great job of finding debug info -- and especially, allowing the user to customize the process. So in my development branch, we're dropping most of the code that searches the "standard directories" for debuginfo, and just relying on drgn to do that. We'll provide some extra logic for some things that may be a bit more specific to Oracle Linux, but lookups in the standard paths will use drgn's standard logic, and thus should work well on any distro without modification.

I'm also really curious, are there any specific helpers or tools you're using out of drgn-tools? I didn't know anybody outside Oracle was using it, and if there are rough edges I can help smooth out (including the debuginfo, which I'll do my best to get the changes out quickly), I would be glad to help.

brenns10 (approved):

brenns10:
Thank you! If you'd like, I can mention you on the PR for the broader changes to debuginfo loading, so you can have a look yourself.

PR #170, opened 2025-06-11
Memstate Module

YassineLr:
integrating memstate module into CoreLens.

### DIFF
--- a/drgn_tools/memstate.py
+++ b/drgn_tools/memstate.py
@@ -1,0 +1,843 @@
+# Copyright (c) 2025, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
+"""
+    Helper to view memory statistics
+"""
+import argparse
+import time
+from collections import defaultdict
+from numbers import Number
+from typing import Any
+from typing import Dict
+from typing import List
+from typing import Tuple
+
+from drgn import Program
+from drgn.helpers.linux.cpumask import for_each_present_cpu
+from drgn.helpers.linux.percpu import per_cpu
+from drgn.helpers.linux.pid import for_each_task
+from drgn.helpers.linux.slab import for_each_slab_cache
+from drgn.helpers.linux.slab import get_slab_cache_aliases
+
+from drgn_tools.buddyinfo import get_per_zone_buddyinfo
+from drgn_tools.corelens import CorelensModule
+from drgn_tools.meminfo import for_each_zone
+from drgn_tools.meminfo import get_active_numa_nodes
+from drgn_tools.meminfo import get_all_meminfo
+from drgn_tools.meminfo import get_total_hugetlb_pages
+from drgn_tools.mm import totalram_pages
+from drgn_tools.numastat import get_per_node_meminfo
+from drgn_tools.slabinfo import get_kmem_cache_slub_info
+from drgn_tools.task import get_command
+from drgn_tools.task import get_pid
+from drgn_tools.task import get_task_rss
+
+
+"""
+    these are common global constants used by oled-memstate.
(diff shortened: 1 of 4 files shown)


drgn_tools/memstate.py
```diff
@@ -0,0 +1,846 @@
...
+    return {
+        "total_memory": total_memory,
+        "free_memory": free_memory,
+        "used_memory": used_memory,
+        "kernel_mm_usage": kernel_mm_usage,
+        "user_space": user_space_memory,
+    }
+
+
+def get_slab_inof(prog: Program):
```
brenns10, line 257: nit: `get_slab_info`
also: this function returns a value, but its return value is not annotated. All function return values should be annotated as well. This applies for all functions.

drgn_tools/memstate.py
```diff
@@ -0,0 +1,846 @@
...
+    """
+    this api is returning a dictionnary that contain the slab cache , it size and their aliases
+    """
+    slabs_infos: dict = defaultdict()
```
brenns10, line 261: The type annotation `dict` is not usable for python < 3.9. So, we use `typing.Dict` (and `typing.List`, etc).

Further, if you're going to annotate it, you should probably include types for the keys and values. So that would be:

```
from numbers import Number
from typing import Dict, List

slab_infos: Dict[str, Dict[str, Number]] = defaultdict()
```

drgn_tools/memstate.py
```diff
@@ -0,0 +1,846 @@
...
+            aliases_lst = infos.get("aliases", [])
+            aliases_str = ", ".join(aliases_lst) if aliases_lst else "(null)"
+            print(f"{slab:<35}{int(infos['size']):>12}    {aliases_str}")
+            displayed = displayed + 1
+    print("")
+    print(">> Total memory used by all slab caches: " f"{total_mem} KB")
+
+
+def calculate_pss(task):
+    """ " """
```
brenns10, line 313: Missing function

drgn_tools/memstate.py
```diff
@@ -0,0 +1,846 @@
...
+    for slab, infos in sslab_info.items():
+        if displayed >= 10 and not verbose:
+            break
+        else:
+            if infos["size"] == 0:
+                break
+            aliases_lst = infos.get("aliases", [])
+            aliases_str = ", ".join(aliases_lst) if aliases_lst else "(null)"
+            print(f"{slab:<35}{int(infos['size']):>12}    {aliases_str}")
+            displayed = displayed + 1
```
brenns10, line 307: Maybe personal preference, but having an `if ... break`, to me I don't think there needs to be an "else" clause. We understand that control flow would only reach outside of the if statement, if the condition were false.

And by avoiding the `else` clause you get smaller indentation, which is useful frequently (though not here specifically).

brenns10 (changes requested):
Here's a quick review. In addition to the comments on the code below, here are a few quick and easy fixes.

(1) Use `git commit -s` (sign-off) for your commits, so they include your "Signed-off-by" tag. See prior commits in history for examples.
(2) Check the following pre-commit findings:
```

drgn_tools/memstate.py:111:25: E221 multiple spaces before operator
drgn_tools/memstate.py:365:27: E226 missing whitespace around arithmetic operator
drgn_tools/memstate.py:365:33: E225 missing whitespace around operator
drgn_tools/memstate.py:365:45: E225 missing whitespace around operator
drgn_tools/memstate.py:383:31: E226 missing whitespace around arithmetic operator
drgn_tools/memstate.py:383:37: E225 missing whitespace around operator
drgn_tools/memstate.py:383:49: E225 missing whitespace around operator
drgn_tools/memstate.py:499:25: E231 missing whitespace after ','
drgn_tools/memstate.py:503:62: E231 missing whitespace after ','
```

brenns10 (approved):

PR #172, opened 2025-06-16
Updated sosreport plugin changes

brenns10:
These are @apalakun's changes (see #155), but with some code style updates, fixes for the spec file, and some simplifications in the code to avoid unnecessary RPM version checks. I have built the RPM and tested `sos report --batch` with vmcores present on OL9, and the corelens report was included in the output -- slick!

### DIFF
--- a/extras/corelens.py
+++ b/extras/corelens.py
@@ -126,7 +126,13 @@ class LibCorelens(Plugin, RedHatPlugin):
             corelens_output_path.mkdir(parents=True, exist_ok=True)
 
             for vmcore_subdir in recent_vmcore_dirs:
-                vmcore_file = vmcore_subdir / "vmcore"
+                try:
+                    vmcore_file = next(vmcore_subdir.glob("*vmcore"))
+                except StopIteration:
+                    error_logs.append(
+                        f"Error: missing vmcore in {vmcore_subdir}"
+                    )
+                    continue
                 corelens_output_file = (
                     corelens_output_path / vmcore_subdir.name
                 )


apalakun (approved):
Changes looks and approved

extras/corelens.py
```diff
@@ -0,0 +1,136 @@
...
+                error_logs.append("No recent vmcore directories found.")
+                self.add_string_as_file("\n".join(error_logs), filename="corelens")
+                return
+
+            archive_tmp_dir = self.archive.get_tmp_dir()
+            corelens_output_path = Path(archive_tmp_dir) / "sos_commands/corelens"
+            corelens_output_path.mkdir(parents=True, exist_ok=True)
+
+            for vmcore_subdir in recent_vmcore_dirs:
+                vmcore_file = vmcore_subdir / "vmcore"
```
biger410: Is file name always "vmcore"? I used to see vmcore name like `$hostname.vmcore` from exadata system.

extras/corelens.py
```diff
@@ -0,0 +1,136 @@
...
+        corelens_binary = Path("/usr/bin/corelens")
+        if corelens_binary.exists() and self.check_version():
+            self.process_recent_vmcores(days)
+        else:
+            corelens_info_header = "Corelens details::"
+            rpm_cmd = "rpm -qa | grep -E 'oled|drgn'"
+            try:
+                corelens_pkg = subprocess.check_output(rpm_cmd, shell=True, universal_newlines=True).decode("utf-8")
+            except Exception:
+                corelens_pkg = "Error fetching package details either oled-tools or drgn package is missing"
```
biger410: What function from oled-tools is required to support this feature?

extras/corelens.py
brenns10: Good point, maybe we should select any file that ends with "vmcore".

brenns10: Updated in the latest commit

biger410 (approved):

PR #175, opened 2025-07-09
Adopt the Module API in drgn-tools

brenns10:
I apologize in advance for the size of this pull request, as well as the relative disorganization of the commits. My design for this changed one or twice during the implementation, and I didn't end up squashing those changes down. The result is that the branch shows the evolution of the design, which is reasonable, but I wouldn't necessarily recommend reviewing individual commits, except to read the messages with rationales.

This adopts the module API and moves our minimum drgn version to 0.0.32 (despite what the branch name says). Since this is the latest, and the last version supporting Python 3.6, we will likely be using this as our minimum for quite a long time. As such, I didn't see any reason to keep support for any older version.

The goals here were:

1. Remove code that reimplemented parts of drgn's debug info finding logic.
2. Make Corelens more reliable for detecting module debuginfo.
3. Get rid of the bespoke debuginfo loading mechanisms for `corelens` and `drgn_tools.cli`, instead relying on drgn debug info finders, and making their behavior more consistent.
4. With the power of debuginfo finders & plugins, add the ability to download and extract OL debuginfo RPMs to all commands, including the drgn command.
5. Remove old, dead code that supported older drgn versions. EG: code upstreamed to drgn.

This adds a drgn plugin, with several debug info finders. It revamps the (hardly ever used) drgn-tools config file. It does modify some elements of the corelens & drgn_tools.cli arguments. It removes the `KernelModule` class. And there are several other major changes. I've written a document that goes into far more detail than I feel comfortable writing in a PR description. I'll share it to those curious.


brenns10:
Ok, finally got all tests passing internal & external. I think this is ready for review.

biger410 (approved):
Looks good.

PR #176, opened 2025-07-10
rds.py: Added RDMA restrack resource usage in rds.py script

PalakTripathi1:
Enhance rds.py to report RDMA restrack resource usage (PD, CQ, QP, CM_ID, MR, CTX, SRQ) per RDS interface. This is based on the xa-for-each traversal of res[] arrays in struct ib_device. Useful for visibility into actual RDMA object usage and correlating with rdma resource show script

### DIFF
--- a/drgn_tools/rds.py
+++ b/drgn_tools/rds.py
@@ -83,15 +83,6 @@ RDMA_CM_STATES = {
     10: "RDMA_CM_DESTROYING",
 }
 
-RDMA_RES_TYPES = {
-    0: "pd",
-    1: "cq",
-    2: "qp",
-    3: "cm_id",
-    4: "mr",
-    5: "ctx",
-    7: "srq",
-}
 # Helpers #
 
 def be64_to_host(prog: drgn.Program, value: int) -> int:
@@ -586,13 +577,10 @@ def rdma_resource_usage(prog: Program, outfile: Optional[str] = None, report: bo
     :param outfile: A file to write the output to.
     :param report: Whether to open file in append mode for report.
     """
-    from drgn.helpers.linux import xa_for_each, list_for_each_entry
-    from drgn import container_of
-
     dev_kset = prog["devices_kset"]
     data = [["Index", "Device", "PD", "CQ", "QP", "CM_ID", "MR", "CTX", "SRQ"]]
     index = 0
-
+    res_types_enum = prog.type("enum rdma_restrack_type")
     for dev in list_for_each_entry("struct device", dev_kset.list.address_of_(), "kobj.entry"):
         try:
             name = dev.kobj.name.string_().decode()
@@ -601,7 +589,8 @@ def rdma_resource_usage(prog: Program, outfile: Optional[str] = None, report: bo
             if not dev_name.startswith("mlx"):
                 continue
             counts = {}
-            for i, res_name in RDMA_RES_TYPES.items():
+            for name, i in res_types_enum.enumerators:
+                res_name = name[len("RDMA_RESTRACK_"):].lower()
                 try:
                     xa = ib_dev.res[i].xa
                     counts[res_name] = sum(1 for _ in xa_for_each(xa.address_of_()))
@@ -1583,7 +1572,7 @@ def report(prog: drgn.Program, outfile: Optional[str] = None) -> None:
     if msg:
         print(msg)
         return None
-        
+
     rds_dev_info(prog, outfile=outfile, report=False)
     rdma_resource_usage(prog, outfile=outfile, report=False)
     rds_sock_info(prog, outfile=outfile, report=True)


drgn_tools/rds.py
```diff
@@ -1527,7 +1616,7 @@ def report(prog: drgn.Program, outfile: Optional[str] = None) -> None:
     if msg:
         print(msg)
         return None
-
+    rds_rdma_resource_usage(prog, outfile=outfile, report=False)  # NEW
```
anandakhoje, line 1619: Please remove the "# NEW" comment here.

drgn_tools/rds.py
```diff
@@ -1403,6 +1431,67 @@ def rds_print_msg_queue(
         )
 
 
+def rds_rdma_resource_usage(prog: Program, outfile: Optional[str] = None, report: bool = False) -> None:
```
anandakhoje, line 1434: This resource usage is not specific to RDS, so I think we should drop rds from the name here.

drgn_tools/rds.py
```diff
@@ -1403,6 +1431,67 @@ def rds_print_msg_queue(
...
+    """
+    Print RDMA restrack resource usage counts per device based on rds_ib_devices
+
+    :param prog: drgn.Program
+    :param outfile: A file to write the output to.
+    :param report: Whether to open file in append mode for report.
+    """
+    from drgn.helpers.linux import xa_for_each, list_for_each_entry
+
+    rds_ib_devices = prog["rds_ib_devices"].address_of_()
```
anandakhoje, line 1444: Ok, it appears we are only considering the RDS devices for measuring resources, instead I am of the opinion we should consider all registered mellanox devices. Can we use "mlx5_ib_dev_list" here?

drgn_tools/rds.py
```diff
@@ -83,8 +83,27 @@
...
+    2: "qp",
+    3: "cm_id",
+    4: "mr",
+    5: "ctx",
+    7: "srq",
+}
 # Helpers #
 
+def get_rdma_resource_counts(ib_dev) -> dict:
+    counts = {name: 0 for name in RDMA_RESTRACK_TYPES.values()}
```
anandakhoje, line 98: I recommend renaming this to "get_rdma_resources()" and "counts to res_count". Meaningful naming helps understand the code better.

drgn_tools/rds.py
```diff
@@ -83,8 +83,27 @@
     10: "RDMA_CM_DESTROYING",
 }
 
+RDMA_RESTRACK_TYPES = {
```
anandakhoje, line 86: s/RDMA_RESTRACK_TYPES/RDMA_RES_TYPES

drgn_tools/rds.py
```diff
@@ -538,6 +557,13 @@ def rds_dev_info(
             "dev_name",
             "node_name",
             "IP (state)",
+            "PD",
```
anandakhoje, line 560: As this measurement should not be specific to RDS, we need to move this out of rds_dev_info.

anandakhoje (changes requested):
Overall, as "rdma res show" tracks all RDMA resources and not only of that being used by RDS devices, this should be more non-RDS specific. Provided comments inline.

drgn_tools/rds.py
anandakhoje: On VMs, the list mlx5_ib_dev_list is empty.
I think we can use the devices_kset list to get all mlx devices then. Need to see how to get ib_device from it.

>>> dev_kset = prog['devices_kset']
>>> list = dev_kset.list
>>> for dev in list_for_each_entry("struct device", list.address_of_(), "kobj.entry"):
...     str(dev.kobj.name)
... 

^ This prints all devices, filter mellanox devices out by comparing name and see if it contains substring "mlx".

PalakTripathi1: Can have ib device by container_of ``ib_dev = container_of(dev, "struct ib_device", "dev")`

drgn_tools/rds.py
PalakTripathi1: It's Done

drgn_tools/rds.py
```diff
@@ -598,6 +578,60 @@ def rds_dev_info(
...
+            name = dev.kobj.name.string_().decode()
+            if "mlx" not in name:
+                continue
+
+            ib_dev = container_of(dev, "struct ib_device", "dev")
+            dev_name = ib_dev.name.string_().decode()
+
+            if not dev_name.startswith("mlx5_"):
+                continue
+
```
anandakhoje, line 607: Looks like the script is capturing rdma resources utilized by mlx5_* devices only. Do we need a different code for mlx4_* devices?

PalakTripathi1: It's Done

anandakhoje (approved):
Looks good to me.

Reviewed-by: Anand Khoje <anand.a.khoje.oracle.com>

drgn_tools/rds.py
```diff
@@ -570,6 +578,55 @@ def rds_dev_info(
...
+def rdma_resource_usage(prog: Program, outfile: Optional[str] = None, report: bool = False) -> None:
+    """
+    Print RDMA restrack resource usage counts for ALL mlx5_* devices, similar to 'rdma res show'
+
+    :param prog: drgn.Program
+    :param outfile: A file to write the output to.
+    :param report: Whether to open file in append mode for report.
+    """
+    from drgn.helpers.linux import xa_for_each, list_for_each_entry
+    from drgn import container_of
```
brenns10, line 590: Please remove these imports. They're already present at the top of the file, and generally we prefer to have them at the top of the file, not within the function body.

drgn_tools/rds.py
```diff
@@ -570,6 +578,55 @@ def rds_dev_info(
...
+
+    for dev in list_for_each_entry("struct device", dev_kset.list.address_of_(), "kobj.entry"):
+        try:
+            name = dev.kobj.name.string_().decode()
+            ib_dev = container_of(dev, "struct ib_device", "dev")
+            dev_name = ib_dev.name.string_().decode()
+            if not dev_name.startswith("mlx"):
+                continue
+            counts = {}
+            for i, res_name in RDMA_RES_TYPES.items():
```
brenns10, line 604: Rather than using `RDMA_RES_TYPES`, you can do the following. First, outside of all of the loops, look up the type:

```python
res_types_enum = prog.type("enum rdma_restrack_type")
```

Then, you can replace this loop with:

```suggestion
            for name, i in res_types_enum.enumerators:
                res_name = name[len("RDMA_RESTRACK_"):].lower()
```

drgn_tools/rds.py
```diff
@@ -1527,8 +1583,9 @@ def report(prog: drgn.Program, outfile: Optional[str] = None) -> None:
     if msg:
         print(msg)
         return None
-
+
```
brenns10, line 1586: nit: Please drop this extra whitespace.

brenns10 (changes requested):
Just a few small changes requested, otherwise this looks quite good, thank you!

brenns10:
Another thing - can you file a bug for these RDS improvements (targetting "next" -- this is also covered in the workflow doc). Then include the "Orabug: " statement in the commit messages? It can be the one Orabug for both this and #177.

drgn_tools/rds.py
PalakTripathi1: It's done, resolved

brenns10 (approved):
So, [pre-commit still reports formatting issues](https://github.com/oracle-samples/drgn-tools/actions/runs/16435743085/job/46494680408?pr=176), but I'm willing to approve this PR and fix-up those myself.

PR #178, opened 2025-07-10
Add vectorinfo module

srivathsa729:
Prints per-cpu and system vector information.

Authored-by: Partha Satapathy partha.satapathy@oracle.com
Co-authored-by: Srivathsa Dara srivathsa.d.dara@oracle.com

### DIFF
--- a/drgn_tools/vectorinfo.py
+++ b/drgn_tools/vectorinfo.py
@@ -1,0 +1,241 @@
+# Copyright (c) 2025, Oracle and/or its affiliates.
+# Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
+"""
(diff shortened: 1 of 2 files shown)


srivathsa729:
Hi Stephen,

Thanks for the review.

Sorry for the delayed response.

I noticed that 'vector_matrix' is not available in the UEK5 kernel, even on the x86_64 architecture. To address this, I’ve handled that specific case within the print_vector_matrix function.

For non-x86_64 architectures, instead of checking twice (once in print_vector_matrix and once in print_vectors), I now doing the architecture check only once in print_vector_info, and simply return early if the architecture is not x86_64.

brenns10 (approved):
I rebased against the latest main branch to get the 6.16 fixes, and now this all look good, thank you.

PR #180, opened 2025-07-22
[stable] rds: fix crash on NULL task

brenns10:
It seems that sometimes the task could be NULL, and in this case we should allow the module to continue executing.

Orabug: 38225232


imran-kn (approved):
LGTM

PR #183, opened 2025-08-14
New corelens module: pstack

brenns10:
This imports the pstack contrib script from drgn, and updates it. I'll end up pushing many of these updates back to drgn as well, but the main things are:

1. It is now tested on aarch64 and works with minimal changes.
2. Tasks that are running in user-mode on CPU can now be unwound and printed correctly.
3. Task state selection is done so that multiple pids, command patterns, or task states can be specified very flexibly.
4. For the "dump" mode, all task data is dumped to a single file, rather than several files per PID.

### DIFF
--- a/drgn_tools/pstack.py
+++ b/drgn_tools/pstack.py
@@ -34,7 +34,6 @@ import sys
 from bisect import bisect_left
 from pathlib import Path
 from typing import Any
-from typing import BinaryIO
 from typing import Dict
 from typing import List
 from typing import Tuple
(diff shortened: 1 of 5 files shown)


brenns10:
This is ready for review. Since most of the functionality can be exercised on a live system via `/proc/kcore`, the tests here actually do exercise it pretty well. The tests verify the register fetching, the dump format, and ensure that the "pstack" operation outputs the same as the "dump + print" operation.

biger410 (approved):
Looks good.

PR #205, opened 2026-02-27
add helpers to dump max inflight io time for one disk and its disk driver

biger410:
Two columns "DRIVER" and "Max Inflight time" are added.

```
>>> from drgn_tools import blockinfo
>>> blockinfo.print_block_devs_info(prog)
MAJOR  GENDISK             NAME   DRIVER       REQUEST_QUEUE       Inflight I/Os  Max Inflight time  Freezed Depth  Usage Counter
8      0xffff99155b391a00  sda    virtio_scsi  0xffff99154b323990  0              0 00:00:00.000     0              1
252    0xffff99154105c000  dm-0   linear       0xffff99154b321cc8  0              0 00:00:00.000     0              1
252    0xffff9915425e2200  dm-1   linear       0xffff9915c967d658  0              0 00:00:00.000     0              1
8      0xffff9915428ad800  sdb    iscsi_tcp    0xffff99154b320998  0              0 00:00:00.000     0              1
8      0xffff991547dfe800  sdc    iscsi_tcp    0xffff99154b320000  24             0 00:00:00.418     0              30
7      0xffff9915c95ba000  loop0  loop         0xffff9915c96d2ff8  110            0 00:00:16.316     0              111
9      0xffff991547e46c00  md0    raid456      0xffff9915c967dff0  0              0 00:00:00.000     0              4
7      0xffff99154611b200  loop1  loop         0xffff9915c96d0998  48             0 00:00:03.658     0              49
7      0xffff991558d6f000  loop2  loop         0xffff9915c967f320  123            0 00:00:09.403     0              125
7      0xffff9915428acc00  loop3  loop         0xffff9915c96d4328  12             0 00:00:03.990     0              13
7      0xffff99154089d400  loop4  loop         0xffff9915c96d0000  32             0 00:00:02.074     0              33
7      0xffff991542f5d400  loop5  loop         0xffff99154b322660  34             0 00:00:32.587     0              38
7      0xffff991548970000  loop6  loop         0xffff9915c96d1330  118            0 00:00:31.792     0              119
7      0xffff9915594d5200  loop7  loop         0xffff9915c9679330  128            0 00:00:36.261     0              131
9      0xffff9915494cf400  md1    raid1        0xffff991592393990  0              0 00:00:00.000     0              1
9      0xffff99154ab67400  md2    raid0        0xffff991592391330  0              0 00:00:00.000     0              1
>>>
```

### DIFF
--- a/drgn_tools/nvme.py
+++ b/drgn_tools/nvme.py
@@ -21,6 +21,23 @@ from drgn_tools.table import print_table
 from drgn_tools.util import enum_name_get
 
 
+def nvme_ctrl(prog: Program, disk: Object) -> Object:
+    diskname = disk.disk_name.string_().decode()
+    if not diskname.startswith("nvme"):
+        return None
+
+    q = disk.queue
+    ns = Object(prog, "struct nvme_ns *", value=q.queuedata.value_())
+    return ns.ctrl
+
+
+def nvme_disk_driver(prog: Program, disk: Object) -> str:
+    ctrl = nvme_ctrl(prog, disk)
+    if not ctrl:
+        return "unknown"
+    return ctrl.ops.module.name.string_().decode()
+
+
 def for_each_nvme_disk(prog: Program) -> Iterable[Object]:
     """
     Return each NVMe device
--- a/.header.txt
+++ b/.header.txt
@@ -1,2 +1,2 @@
-# Copyright (c) 2025, Oracle and/or its affiliates.
+# Copyright (c) 2026, Oracle and/or its affiliates.
 # Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl/
--- a/drgn_tools/block.py
+++ b/drgn_tools/block.py
@@ -5,7 +5,6 @@ Helpers for block layers.
 
 No kernel modules are required since uek built in all io schduler modules.
 """
-import argparse
 from typing import Iterable
 from typing import Tuple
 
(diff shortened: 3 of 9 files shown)


drgn_tools/nvme.py
```diff
@@ -21,6 +21,23 @@
...
 
 
+def nvme_ctrl(prog: Program, disk: Object) -> Object:
+    diskname = disk.disk_name.string_().decode()
+    if not diskname.startswith("nvme"):
+        return None
+
+    q = disk.queue
+    ns = Object(prog, "struct nvme_ns *", value=q.queuedata.value_())
+    return ns.ctrl
```
brenns10, line 31: Is this the best way to get the `nvme_ns`? Compared to consulting `disk->fops` and `disk->private_data`?

brenns10 (approved):
This all looks pretty reasonable to me. One question on the NVME code but I'd defer to you on it.

More broadly - are the disk names the best way to detect what subsystem is used for the disk? Seems to me like `struct gendisk` has `fops` which might point to a specific operations struct, which seems like a more canonical way of detecting these things.

Again, I'd defer to your judgment there, so consider this approved if these aren't really valid concerns.

drgn_tools/nvme.py
biger410: Make sense. I will fix these two and repost. thanks for the review.

biger410: Actually "disk.private_data" could be `nvme_ns` or `nvme_ns_head`, and "disk.fops" could be `nvme_bdev_ops` or `nvme_ns_head_ops`, depending on whether it's nvme multipath device. Also "disk.fops " is `nvme_fops` in older kernel. We can handle these, but it will take more code and it may change in future kernel. Current way to detect the disk type by name didn't look "reliable", but actually this name conversion has been used since the beginning of linux and doesn't look like it will change in future. I think i will keep current approach.

PR #209, opened 2026-04-06
Fix "ValueError: 2950289701 is not a valid MemSlotFlag" Error

SiddhiK29:


brenns10 (changes requested):
You've encountered the one gap in our "flags" helper code! Sorry :(

There are flags which are implemented as kernel enums, and flags which are just preprocessor constants that we need to encode as Python `enum.Enum`. Separate from this distinction, sometimes the enum/macro value is a bit number, and other times it is the actual bit value.

For kernel enums, we can use [decode_flags](https://drgn.readthedocs.io/en/latest/helpers.html#drgn.helpers.common.format.decode_flags) and specify `bit_number`. For python enums, we have `drgn_tools.util.BitNumberFlag` but that's only for enums that specify bit numbers. I think it would be good to have an equivalent class, `BitFlag`, which does the translation for enums which are not bit numbers. This would be used for `MemSlotFlag()`. Something like this:
```python
# in drgn_tools/util.py below BitNumberFlags
class BitFlags(IntEnum):
    @classmethod
    def decode(cls, value: int) -> str:
        names = []
        for bit in cls:
            if bit & value:
                names.append(bit.name)
                value &= ~(bit)
        if not names or value:
            names.append(f"0x{value:x}")
        return "|".join(names)
```

And then, `MemSlotFlag` can inherit from `BitFlags` and you decode to a string like so:
```
print(MemSlotFlag.decode(5))
```
This way you do not need `NONE = 0` in the declaration of the flags, and you'll be able to support any combination of flags. Sorry this was not readily available in drgn-tools already!

brenns10 (approved):
This approach works too, thanks!

PR #210, opened 2026-04-14
runq: dump task's schedule delay and rt queue throttling status

biger410:
This will help diagnostic scheduling related issues.


brenns10 (approved):
Both look good to me!

PR #211, opened 2026-04-20
rds: fix rds_conn_cpu_info() on UEK5/6

brenns10:
The preferred_send_cpu and preferred_recv_cpu fields don't exist on older UEK5 & UEK6 versions, causing crashes on our historic vmcores. It seems the original field was originally i_irq_local_cpu, which became i_preferred_cpu, which was then split into the send and recv CPU fields, and then the recv_sibling was added.

Since RDS on UEK is developed mostly in tandem across all UEK releases, there's not a single upstream commit or UEK release where things change. Instead they seem to change mid-release. Rather than trying to detect all the possible combinations, let's just print all the fields we find.

Fixes: 0c3fa32 ("rds: Added a verbose option to extract additional debug information")

### DIFF
--- a/drgn_tools/rds.py
+++ b/drgn_tools/rds.py
@@ -1755,21 +1755,45 @@ def rds_conn_cpu_info(
         return None
 
     index = -1
-    conn_list = Table(
-        [
-            " ",  # index
-            "rds_conn",
-            "ib_conn",
-            "Conn Path",
-            "ToS",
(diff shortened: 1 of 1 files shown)


brenns10:
This fixes the vmcore tests on Gitlab which have been failing for a bit.

biger410 (approved):

PR #212, opened 2026-05-24
smp: fix smp module for uek7 and later kernels.

imran-kn:
In UEK7 and later kernels CSD framework is being used for task wakeup. So now in a CPU's call_single_queue we will have CSDs for task wakeup as well and these will not have source and destination fields initialized.

Report task wakeup CSDs separately and avoid these CSDs being treated as ones with both source and destination as cpu 0.

Orabug: 39429804

### DIFF
--- a/drgn_tools/smp.py
+++ b/drgn_tools/smp.py
@@ -6,6 +6,7 @@ from typing import Iterator
 from typing import Tuple
 
 import drgn
+from drgn import container_of
 from drgn import Object
 from drgn import Program
 from drgn import Type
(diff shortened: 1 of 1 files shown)


drgn_tools/smp.py
```diff
@@ -288,6 +293,28 @@ def dump_smp_ipi_state(prog: drgn.Program) -> None:
...
             else:
                 src = csd.node.src.value_()
+                # Linux kernel commit 90b5363acd47 ("sched: Clean up scheduler_ipi()")(in v5.10)
+                # started using CSD framework for remote wakeups.
+                # Report tasks in the wake list but don't report source and destination CPUs for
+                # such CSDs because these fields are not used in this case.
+                if (
+                    csd.node.u_flags.value_() & _CSD_FLAG_TYPE_MASK
+                    == _CSD_TYPE_TTWU
+                ):
```
brenns10, line 303: For what it's worth, these values can be gotten with `prog.constant("CSD_TYPE_TTWU")` and `prog.constant("CSD_FLAG_TYPE_MASK")`. That said, I'm not going to block merging this, because I think it's a lot more complicated to look it up (and catch the error when the value is not present), and these flags are pretty stable.

brenns10 (approved):

PR #213, opened 2026-06-01
New test runner

brenns10:
Introduce a new test runner, `testing.vm`, which will replace both the `testing.litevm` and `testing.heavyvm` runners, while adding more features and being overall... better.

The new test runner has the following benefits:
- It runs tests in an Oracle Linux rootfs. This means the tests run in a real OL userspace and can use the official `drgn` RPMs in our tests. (Compared to the old "litevm" runner, which ran in the host's rootfs.)
- The rootfs is built with `dnf --installroot` in a container, which is much faster than scripting the ISO installer, and easier to update than the prebuilt cloud VM images.
- Since the rootfs is manually built and part of the host filesystem, we can install just what we need, keeping the size small. And since we can share the rootfs for each test associated test kernel, we have less duplication. My rootfs directory is less than 2 GiB for all three combined.
- Since the rootfs is not contained in an image, it's easy to run maintainence commands (e.g. installing new packages, compiling kernel modules) in a chroot rather than spinning up a full VM.
- As a direct consequence, we can now build and insert a kernel module for our VM tests.
- The runner now supports testing against Red Hat kernels (RHCK) -- including with the test kernel module.
- Tests run on Oracle Linux host with only standard RPMs - just be sure to use the latest KVM utils stream.
- Tests boot and run much faster! My laptop can run both DWARF + CTF test for OL8/UEK6 in around 15 seconds -- and this requires booting twice.

Some other notes:
- This is a large amount of addition without many changes or deletions. The idea is to add this new framework in parallel to the others. I'll remove the others (`testing.litevm`, `testing.heavyvm`) once this is stable on Github CI.
- I have some follow-ups that I will tackle later based on priority:
  - Make it possible to run tests against specific Python versions in each rootfs. (OL8: python 3.6, 3.12, OL9: 3.9, 
3.12, 3.14 soon, OL10: 3.12, 3.14 soon).
  - Enable running vmcore tests in the chroot for each OL version. This ensures that vmcore tests also use the real bits from our RPMs.
  - Make it easy to drop-in new drgn/drgn-tools RPMs to test them (replacing my ad-hoc testing workflow for testing new RPM releases).
  - Expand support to aarch64 and add this to Gitlab.

There is a known failure on UEK-NEXT due to changes in the slab allocator. Imran is currently taking a look at it, but I don't think this should block merging this.

### DIFF
--- a/.github/workflows/litevm.yml
+++ b/.github/workflows/litevm.yml
@@ -25,48 +25,37 @@ jobs:
       - name: Run pre-commit hooks
         run: pre-commit run --all-files --show-diff-on-failure
   test:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     strategy:
       matrix:
-        # Our minimum supported version of Python is 3.6, used by Oracle Linux 7
-        # & 8. Ideally we would run tests on that, along with Python 3.9 for
-        # Oracle Linux 9, and maybe Python 3.12 for an upcoming Oracle Linux 10.
-        # However, practicality rules here. The binutils provided in Ubuntu
-        # 20.04 is not recent enough for our libctf usage, but 20.04 is the only
-        # remaining Ubuntu image with Python 3.6 available in Github actions.
-        # So we have to eliminate Python 3.6 from our test matrix. Other CI
-        # tests do run on Python 3.6, and there is the "vermin" pre-commit hook
-        # which detects code incompatible with 3.6.
-        python-minor: ["9", "12"]
+        ol-version: [8, 9, 10]
       fail-fast: false
     steps:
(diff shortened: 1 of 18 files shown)


biger410:
Was the new runner running inside other OS, or qemu with oracle linux?

brenns10:
The new runner is QEMU with Oracle Linux user space. 
Previously it was:
- QEMU with whatever user space the host provides (Ubuntu for GitHub CI). This was the litevm tests.
- QEMU with a Oracle Linux user space (as a qcow2 image that needed rebuilding every so often). This was the heavyvm tests on Gitlab.

We should be able to replace both with the new runner.

biger410 (approved):

PR #216, opened 2026-06-09
Minor bug fixes and improvements before the release.

brenns10:
This is a bit of a grab-bag of new features and fixes:
- Adds the `%corelens` command for drgn
- Corrects a few bugs that I had failed to address from bugdb
- Add ksplice cold-patch detection for the debuginfo loader.
- Add a fix and workaround for cases where corelens tries to extract debuginfo for modules that aren't in the debuginfo RPM.
- Include the `dnf` package (and yum repositories) in the rootfs.

The live test cases should be passing at this point. I'll need to audit the vmcore test suite to ensure it's still passing. After that, I'll create the PR with packaging changes and the release.

### DIFF
--- a/drgn_tools/cpuinfo.py
+++ b/drgn_tools/cpuinfo.py
@@ -15,6 +15,7 @@ from drgn import sizeof
 from drgn import TypeKind
 from drgn.helpers.linux.bitops import for_each_set_bit
 from drgn.helpers.linux.bitops import test_bit
+from drgn.helpers.linux.cpumask import for_each_online_cpu
 from drgn.helpers.linux.percpu import per_cpu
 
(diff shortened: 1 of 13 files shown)


drgn_tools/runq.py
imran-kn, line 1: LGTM.

drgn_tools/cpuinfo.py
imran-kn, line 1: Looks like commit message has some typos: "but no the number of inline ones. Sum the online CPU mask instead." should be "but not the number of online ones. Sum the online CPU mask instead."

Other than this, LGTM.

imran-kn (approved):
LGTM.

PR #217, opened 2026-07-01
Release v2.3.0

brenns10:


tests/test_lock.py
```diff
@@ -12,11 +12,9 @@
 from tests import skip_live
 from tests import skip_unless_have_kmod
 from tests import skip_unless_vmcore
-from tests import skip_vmcore
 
 
 class TestLock(DrgnToolsTestCase):
-    @skip_vmcore("*uek4*")
```
imran-kn, line 19: waitq change is good. Just one question for my understanding, from the earlier code it seems that we were skipping the uek4 vmcores using this directive, so why are we not continuing like that

imran-kn (dismissed):
LGTM. I had one small query but its not a blocker

tests/test_lock.py
brenns10: The skip was added pretty recently, due to the change adding waitqueue support for the lock module. I removed the skip because of the changes in [lock: add UEK4 support for wait queues](https://github.com/oracle-samples/drgn-tools/pull/217/commits/9c9047b1ff76a3534b0d9bec0c4ef818717cab80). Once the renames are handled, the test runs fine on UEK4.

brenns10:
Thanks @imran-kn! I went ahead and addressed one comment from the userspace review (switching `Requires(post)` and `Requires(postun)` to depend on `alternatives` rather than `%{_sbindir}/alternatives`), and added a reviewed-by tag. When you have a moment, can you refresh the approval so I can merge? Thanks again.

imran-kn (approved):
LGTM

brenns10:
Thank you!

PR #218, opened 2026-07-21
mlx5: add Corelens device and queue diagnostics

aeschyl:
## Summary

- Add an mlx5 Corelens module for device, network, RDMA, queue, CQ, EQ, and QP state.
- Support summary, full, and JSON output with filters and bounded descriptor dumps.
- Add package integration, API documentation, and tests.

## Testing

- Verified summary runs with CTF and DWARF vmcores.

Orabug: 39751588


drgn_tools/mlx5_support/compat.py
```diff
@@ -0,0 +1,245 @@
...
+from drgn import ObjectAbsentError
+from drgn import OutOfBoundsError
+from drgn import Program
+from drgn import sizeof
+from drgn import TypeKind
+
+try:
+    from drgn.helpers.linux.irq import irq_to_desc
+except ImportError:  # drgn 0.0.32 does not provide the IRQ helpers.
+    irq_to_desc = None  # type: ignore
```
brenns10: drgn-tools has `drgn_tools.irq.irq_to_desc()`, please use that and delete this.

drgn_tools/mlx5_support/compat.py
```diff
@@ -0,0 +1,245 @@
...
+_MEMBER_ERRORS = (
+    LookupError,
+    FaultError,
+    ObjectAbsentError,
+    TypeError,
+)
+_VALUE_ERRORS = (FaultError, ObjectAbsentError, TypeError, ValueError)
+
+
+def _safe_iter(
```
brenns10: We can discuss on an individual basis if you'd like, but I think this entire `drgn_tools.mlx5_support.compat` module should be deleted.

Most of these "safe" helpers are predicated on the idea that suppressing exceptions makes the code safer or more reliable. And yes, they will reduce the number of exceptions that the code raises. But that actually runs counter to the goal of creating reliable, accurate software. Exceptions such as faults, lookup errors, attribute errors -- these are all violations of the assumptions that the programmer made when they wrote the code. You write the code with an understanding of some kernel data structures, so when you access `struct foo.bar` and the member does not exist, this means your assumption was violated.

That's a signal to us that we need to rework the code. If we suppress all these signals, then it will appear to us as if the code runs fine, but when we actually go to use the result, it may be full of "None" or worse, silently incorrect data because we suppressed errors.

Of course there's a balance to be kept. For example, you may want to wrap the entire section of code that prints port information with a try/except block, so that even if an unexpected failure happens there, you could continue on to print queue information. But even then, you want to surface the error, not simply hide it away.

Another reason this is really important is that a lot of our tests are actually more like "smoke tests". They don't have a lot of assertions, because it's very difficult to write an accurate assertion for their results. The main purpose of these smoke tests is to ensure that the code simply runs: that the assumptions made in the code are still valid. If we suppress them all, then we also lose that value from testing.

drgn_tools/mlx5_support/defs.py
```diff
@@ -0,0 +1,266 @@
...
+"""Static definitions for the mlx5 Corelens report."""
+import re
+from typing import List
+from typing import Tuple
+
+
+_PCI_BDF_RE = re.compile(
+    r"^[0-9a-fA-F]{4}:[0-9a-fA-F]{2}:[0-9a-fA-F]{2}\.[0-7]$"
+)
+_MLX5_DEVICE_STATE = {1: "UP", 2: "INTERNAL_ERROR"}
```
brenns10: One benefit of using drgn is that it understands the types of the program. Enumerations are types, and as such, we can rely on drgn being able to tell us the values and names of the enum. Take `MLX_DEVICE_STATE_UP`, for example. Drgn can look it up the value of the constant, and it can also provide us the enum type and all its values.
```python
>>> prog.constant("MLX5_DEVICE_STATE_UP")
(enum mlx5_device_state)MLX5_DEVICE_STATE_UP
>>> prog.constant("MLX5_DEVICE_STATE_UP").value_()
1

>>> prog.type("enum mlx5_device_state")
enum mlx5_device_state {
        MLX5_DEVICE_STATE_UP = 1,
        MLX5_DEVICE_STATE_INTERNAL_ERROR = 2,
}
```

This means that we don't need to hard-code most constants. (You may see some constants hard-coded in drgn/drgn-tools anyway. These are because preprocessor constants, e.g. `#define RANDOM 4`, are not types, so drgn cannot find them.)

So my recommendation for this module is to delete all of the hard-coded constants. When you need a constant value, you can use [Program.constant](https://drgn.readthedocs.io/en/latest/api_reference.html#drgn.Program.constant) to look it up. And if you need to know all the constant values in order to create a flags string like `UP|INTERNAL_ERROR`, then you can use [decode_enum_type_flags](https://drgn.readthedocs.io/en/latest/helpers.html#drgn.helpers.common.format.decode_enum_type_flags). 

There are several benefits to doing it this way:

1. If future versions of the kernel change the definitions of the constants, your code won't need to be changed. (This is probably unlikely, especially if the constants are shared with firmware/devices...)
2. If new constants are added to the enumerations, your code will automatically become aware of it, without you needing to maintain the list.
3. It's less code to read, write, and maintain!

drgn_tools/mlx5_support/decode.py
```diff
@@ -0,0 +1,167 @@
...
+
+from .defs import _MLX5_CQ_ERROR_SYNDROME
+from .defs import _MLX5_CQE_BYTE_COUNT_REQ_OPCODES
+from .defs import _MLX5_CQE_ERROR_OPCODES
+from .defs import _MLX5_CQE_OPCODE
+from .defs import _MLX5_CQE_RESPONDER_OPCODES
+from .defs import _MLX5_CQE_SYNDROME
+from .defs import _MLX5_EQE_QP_SRQ_TYPES
+from .defs import _MLX5_EQE_TYPE
+from .defs import _MLX5_WQE_OPCODE
```
brenns10: Building on my comment regarding hard-coding constants, I would suggest that you remove these and then the functions in this module would either take a `Program` as an argument, or they could take a `Type` or dictionary of the constants they care about.

drgn_tools/mlx5_support/decode.py
```diff
@@ -0,0 +1,167 @@
...
+
+def _enum_table_label(
+    value: Optional[int], names: Dict[int, str]
+) -> Optional[str]:
+    if value is None:
+        return None
+    return names.get(value) or f"unknown({value})"
+
+
+def _decode_cqe(raw: bytes) -> Dict[str, Any]:
```
brenns10: I'm just curious, is there a struct corresponding to this format you're decoding?

I'm assuming these are hardware ring buffer entries and so the format is not going to change. But if there's a struct, you can let drgn do most of the parsing for you by creating an object with that as the type. For example:

```python
>>> my_object = drgn.Object(prog, prog.type("struct foobar"), value=my_bytes)
>>> my_object.field
```
The only thing you'd need to care about is byte ordering issues, which I can help you with if it comes down to it.

drgn_tools/mlx5_support/collect_device.py
```diff
@@ -0,0 +1,1054 @@
...
+    _want_cqs: bool
+    _want_eqs: bool
+    _want_qps: bool
+    _cqs: Dict[Any, Any]
+    _eqs: Dict[Any, Any]
+    _qps: Dict[Any, Any]
+    _mlx5_ib_devices_cache: Optional[List[Object]]
+    _collect_channels: Callable[..., Any]
+    _iter_walk_limited: Callable[..., Iterable[Any]]
+    _warn: Callable[[str], None]
```
brenns10: Why is this a "mixin" class, as opposed to being implemented directly on `drgn_tools.mlx5.Mlx5Collector`? Generally mixins are defined when you want to add behavior to multiple classes which have different base classes. Here it's used just once.

The member type annotations here are an example of why this doesn't make much sense. The members annotated here are all defined and set by `Mlx5Collector`. So this couldn't really be "mixed in" with other classes and used in any meaningful way.

I'd recommend consolidating everything into one class, (if a class is even necessary).

drgn_tools/mlx5_support/collect_device.py
```diff
@@ -0,0 +1,1054 @@
...
+                    self.prog["init_net"].dev_base_head.address_of_(),
+                    "dev_list",
+                ),
+                self._warn,
+                "iterating netdevs",
+            ),
+            "net_device discovery",
+        ):
+            saw_netdev = True
+            name = _netdev_name(netdev)
```
brenns10: Can you instead use `drgn_tools.crash_net.for_each_netdev`?

```suggestion
        for name, netdev in for_each_netdev(prog["init_net"].address_of_()):
            saw_netdev = True
```

(It would be even better to use `for_each_netdev()` from drgn, but unfortunately this was added in 0.1.0 and we must be compatible with 0.0.32 for now)

drgn_tools/mlx5_support/collect_device.py
```diff
@@ -0,0 +1,1054 @@
...
+            if self.args.netdev and name != self.args.netdev:
+                continue
+
+            driver = _driver_name_from_netdev(netdev)
+            priv = _mlx5e_priv_from_netdev(netdev)
+            mdev = _safe_member(priv, "mdev")
+            if _is_null(mdev):
+                mdev = None
+            if not _looks_like_mlx5_netdev(netdev, driver, priv, mdev):
+                continue
```
brenns10: Could we simplify this and get rid of the extra, mostly redundant checks?

```suggestion
            if not (_symbol_for_addr(netdev.netdev_ops) or "").startswith("mlx5"):
                continue
```

I may be wrong, but there is usually one reasonably simple way to determine whether a device is related to a driver, that shouldn't involve the host of various strategies that `_looks_like_mlx5_netdev()` does.

drgn_tools/mlx5_support/collect_device.py
```diff
@@ -0,0 +1,1054 @@
...
+        "mdev": _hex(mdev_addr),
+        "mdev_present": mdev is not None,
+        "netdevs": [],
+        "summary": {},
+        "health": {},
+        "capabilities": {},
+        "counts": {},
+        "discovery": {"via": [source], **discovery},
+        "_mdev_obj": mdev,
+    }
```
brenns10: I haven't done a full read of the code yet, but if this device record object is supposed to be used in many different places, it should be something like a class or namedtuple. This can be used in type annotations which helps type checkers as well as human and LLM readers of the code. A `Dict[str, Any]` is practically useless as a type annotation, as it could be anything.

I'd also argue that if you don't have a `mdev` or `mdev_addr`, then what's the point in having a device record at all? Why should this support an `Optional[]` for etiher `mdev` or `mdev_addr`?

brenns10 (changes requested):
I have mainly started reading the ancillary modules: `drgn_tools.mlx5_support.*`. I jumped around within them and flagged a few issues, but I was finding that to be unhelpful for understanding the code. So I pivoted and started from the CorelensModule and followed the logic of device discovery.

From what I can see, this PR discovers mlx5 devices in three different ways: (1) by enumerating netdevs and finding ones that look like mlx5 devices, (2) by enumerating RDMA devices and finding mlx5 devices, and (3) by trying global variables that might look like a global list. I think we can make the third one work universally by using a global list. Here's a minimal example of what I mean:

```python
from typing import Iterator

from drgn import cast, container_of, NULL, Object, Program
from drgn.helpers.linux.list import list_for_each_entry
from drgn.helpers.common.format import escape_ascii_string

from drgn_tools.table import Table


def mlx5_netdev(mdev: Object) -> Object:
    """
    :param mdev: object of type ``struct mlx5_core_dev *``
    :returns: the corresponding ``struct net_device *`` or NULL
    """
    return mdev.mlx5e_res.uplink_netdev


def mlx5_core_ib_device(mdev: Object) -> Object:
    """
    Return the primary RDMA device for the MLX5 device. This does not include
    any special secondary devices.
    :param mdev: object of type ``struct mlx5_core_dev *``
    :returns: the corresponding ``struct ib_device *`` or NULL
    """
    prog = mdev.prog_
    MLX5_INTERFACE_PROTOCOL_IB = prog.constant("MLX5_INTERFACE_PROTOCOL_IB")
    ib_adev = mdev.priv.adev[MLX5_INTERFACE_PROTOCOL_IB]
    if not ib_adev:
        return NULL(prog, "struct ib_device *")

    mlx5_ib = cast("struct mlx5_ib_dev *", ib_adev.adev.dev.driver_data)
    if not mlx5_ib:
        return NULL(prog, "struct ib_device *")

    return mlx5_ib.ib_dev.address_of_()


def for_each_mlx5_core_dev(prog: Program) -> Iterator[Object]:
    """
    Iterate over all MLX5 devices on the system.
    :returns: an iterator of ``struct mlx5_core_dev *``
    """
    mlx5_driver = prog["mlx5_core_driver"]

    for knode in list_for_each_entry(
        "struct klist_node",
        mlx5_driver.driver.p.klist_devices.k_list.address_of_(),
        "n_node",
    ):
        dev_priv = container_of(
            knode, "struct device_private", "knode_driver"
        )
        pdev = container_of(dev_priv.device, "struct pci_dev", "dev")
        yield cast("struct mlx5_core_dev *", pdev.dev.driver_data)


def mlx5_report(prog: Program) -> None:
    table = Table(["MLX5_CORE_DEV", "NETDEV", "RDMA_DEV"])
    for mdev in for_each_mlx5_core_dev(prog):
        netdev = mlx5_netdev(mdev)
        netdev_name = escape_ascii_string(netdev.name.string_()) if netdev else "-"
        ibdev = mlx5_core_ib_device(mdev)
        ibdev_name = escape_ascii_string(ibdev.name.string_()) if ibdev else "-"
        table.row(hex(mdev), netdev_name, ibdev_name)
    table.write()


if __name__ == "__main__":
    mlx5_report(prog)

```

drgn_tools/mlx5.py
```diff
@@ -0,0 +1,2477 @@
...
+            OrderedDict()
+        )
+
+    def _discover_devices(self) -> List[DeviceRecord]:
+        devices: List[DeviceRecord] = []
+        saw_netdev = False
+
+        for mdev in collect_device.for_each_mlx5_core_dev(self.prog):
+            if not mdev:
+                raise ValueError("mlx5 driver device has no mlx5_core_dev")
```
brenns10: I don't believe `for_each_mlx5_core_dev()` can ever yield a NULL/None entry, so you can safely get rid of this check.

drgn_tools/mlx5.py
```diff
@@ -0,0 +1,2477 @@
...
+            device = DeviceRecord(mdev)
+            ibdev = collect_device.mlx5_core_ib_device(mdev)
+            if ibdev:
+                ib_mdev = ibdev.mdev
+                if not ib_mdev:
+                    raise ValueError("mlx5_ib_dev.mdev is NULL")
+                if int(ib_mdev) != mdev_addr:
+                    raise ValueError(
+                        "mlx5_ib_dev.mdev does not match its core device"
+                    )
```
brenns10: I agree that the `mdev` field should be non-NULL and should point back to the original `mdev`. That seems to be a mandatory field and the only correct value. But unless we have a reason to expect it to be wrong, or if the code here relies on the assumption, the check isn't that useful.

So my gut reaction is to delete the check.

If there's a reason to have the check that I'm missing, then it would still be better to document it as an assertion, e.g.: `assert int(ibdev.mdev) == mdev_addr`. This is nicer because it's shorter (so easier to read/write), and also because assertions tend to communicate that a core assumption has been violated and needs to be fixed.

drgn_tools/mlx5.py
```diff
@@ -0,0 +1,2477 @@
...
+                raise ValueError(
+                    f"{name}: mlx5e_priv.mdev does not match its core device"
+                )
+            device.netdev = {
+                "name": name,
+                "summary": {},
+                "channels": [],
+                "_priv_obj": priv,
+            }
+            devices.append(device)
```
brenns10: Rather than having an odd continue statement in the middle of the loop, it may make more sense to structure this the same way as you structure the `ibdev` section, where the condition is `if netdev` and all the logic for handling the netdev lives under there. Plus, for similar reasons as I suggested for the ibdev, let's get rid of the mdev check, unless there's a reason this should be explicitly handled.
```suggestion
            if netdev:
                priv = netdev_priv(netdev, "struct mlx5e_priv")
                device.netdev = {
                    "name": name,
                    "summary": {},
                    "channels": [],
                    "_priv_obj": priv,
                }
            devices.append(device)
```

drgn_tools/mlx5.py
```diff
@@ -0,0 +1,2477 @@
...
+            if not mdev:
+                raise ValueError("mlx5 driver device has no mlx5_core_dev")
+            mdev_addr = int(mdev)
+            netdev = mdev.mlx5e_res.uplink_netdev
+            name = None
+            if netdev:
+                name = escape_ascii_string(netdev.name.string_())
+                saw_netdev = True
+            if self.args.netdev and name != self.args.netdev:
+                continue
```
brenns10: Can you move this logic to live in `_filter_devices()` with the rest of the filtering logic?

drgn_tools/mlx5.py
```diff
@@ -0,0 +1,2477 @@
...
+
+    # Channels, Ethernet queues, and their CQs
+
+    def _constant(self, name: str) -> Object:
+        constant = self._constants.get(name)
+        if constant is None:
+            constant = self.prog.constant(name)
+            self._constants[name] = constant
+        return constant
+
```
brenns10: I wouldn't bother with this caching. You can just use `prog.constant()` directly. Drgn maintains an internal hash table mapping names to objects once they're accessed, so this doesn't change much.

drgn_tools/mlx5.py
```diff
@@ -0,0 +1,2477 @@
...
+                    channel_total += 1
+                    queue_total += ptp_queues
+        return channel_total, queue_total
+
+    def _iter_channels(self, channels: Object) -> Iterator[Object]:
+        count = min(int(channels.num), defs.MAX_CHANNELS)
+        for index in range(count):
+            channel = channels.c[index]
+            if channel:
+                yield channel
```
brenns10: This should probably be a proper helper function at the module level, not a class method. Something like:

```python
def mlx5_iter_channels(dev: Object) -> Iterator[Object]:
    """
    Iterate over all ``struct mlx5e_channel *`` from a ``struct mlx5_core_dev *``
    """
    netdev = mlx5_netdev(dev)
    if not netdev:
        return
    channels = netdev_priv(netdev, "struct mlx5e_priv").channels
    for i in range(channels.num):
        channel = channels.c[i]
        if channel:
            yield channel
```

drgn_tools/mlx5.py
```diff
@@ -0,0 +1,2477 @@
...
+
+    def _iter_channel_queues(
+        self, channel: Object
+    ) -> Iterator[Tuple[str, Object, str, Optional[int]]]:
```
brenns10: This should also be a module-level helper rather than a class method, and it should document what each tuple entry represents.

EG:
```python
def mlx5_iter_channel_queues(channel: Object) -> Iterator[...]:
    """
    Given a ``struct mlx5_channel *`` yield all related queues.
    Each queue is represented as a tuple with elements: ...
    """
```

drgn_tools/mlx5_support/render.py
```diff
@@ -0,0 +1,810 @@
...
+
+from . import selection
+from .defs import MAX_DEFAULT_DESCRIPTOR_ENTRIES
+from .format import _display
+from .format import _short_struct
+from drgn_tools.table import Table
+
+DUMP_SEPARATOR = "-" * 79
+
+_DESCRIPTOR_COLUMNS = {
```
brenns10, line 23: Let's go ahead and do an audit of column fields and see which ones are actually getting used and exercised in the various vmcores we have. Then drop the fields we don't have enough data to accurately test, and try to remove the related extraction code for them.

tests/test_mlx5.py
```diff
@@ -0,0 +1,254 @@
...
+
+    def read(self, address, length):
+        for base, data in self.blocks.items():
+            offset = address - base
+            if 0 <= offset and offset + length <= len(data):
+                return data[offset : offset + length]
+        raise FaultError("test memory fault", address)
+
+
+class TestMlx5(unittest.TestCase):
```
brenns10, line 128: I don't have any specific issue with the unit tests below. But the most important tests we want are:

1. mlx5 smoke test that runs on any vmcore (skip live kernels), and if the mlx5_core module is loaded, it runs the mlx5 report and prints it to stdout. This just verifies that the report does not crash.
2. For specific vmcores we can record a known good output and commit it to a file in git, then assert that the output matches.

brenns10 (approved):
OK, looks like we've addressed the major concerns. We have smoke tests in-place and passing on most vmcores. I'll merge this, with a few follow-up items in mind:

1. There are some failures on older UEK6 and UEK5 vmcores. I already included a fix for one. I'll probably clean those up next week.
2. Depending on how things go, I may end up merging the `mlx5_support` package into `mlx5.py` to reduce the footprint further.

